PR #5, opened 2024-07-24
Add / GhostCMS Migrator

ronchambers:
## GhostCMS Migrator copied from Newspack Custom Content Migrator

The primary logic that migrates a GhostCMS JSON file was unchanged from Newspack Custom Content Migrator, but what did change was:

* A [documentation file](https://github.com/Automattic/newspack-migration-tools/blob/f0c6bd603e6590c4bf11a4b80a2f34e7dcc55e87/docs/GhostCMS.md) was added to the `docs/` folder.
* The migrator was divided into a Migrator file and a Helper file.
* The migrator was added to the list of `cli_commands`.
* The way the Migrator/Helper logs to files and cli was updated. 
* Tests were added along with a [json fixture file](https://github.com/Automattic/newspack-migration-tools/blob/f0c6bd603e6590c4bf11a4b80a2f34e7dcc55e87/tests/fixtures/ghostcms.json).

## How to run:

```
wp newspack-migration-tools ghostcms-import --json-file=ghostcms.json --ghost-url=https://newspack.com/ --default-user-id=1
```

## Fatal Conflicts:

If the Newspack Plugin is also active, and the following error has encountered:

```
Error: CoAuthorsPlusHelper construct threw exception: CoAuthors Plus is not installed or active. --> /newspack-repos/newspack-custom-content-migrator/dev/newspack-migration-tools/src/Logic/GhostCMSHelper.php:500
```

Please do [this fix](https://github.com/Automattic/newspack-migration-tools/issues/41):

`wp config set NEWSPACK_ENABLE_CAP_GUEST_AUTHORS true --raw --type=constant`


ronchambers (commented):
I've itemized each piece of the PR here in this review.  Please reply or mark each "resolve conversation" as you see fit.

tests/Command/TestGhostCMSMigrator.php
```diff
@@ -0,0 +1,36 @@
...
+ * Class TestGhostCMSMigrator
+ *
+ * @package newspack-migration-tools
+ */
+class TestGhostCMSMigrator extends WP_UnitTestCase {
+
+	/**
+	 * Test GhostCMS import CLI command
+	 */
+	public function test_cmd_ghostcms_import() {
```
naxoc: Would it be possible to test smaller units than the entire import?

The GhostHelper does most of the heavy lifting, I don't think GhostCMSMigrator needs a test at all. What do you think?

src/Logic/GhostCMSHelper.php
```diff
@@ -0,0 +1,775 @@
...
+
+	/**
+	 * Log wrapper function incase logging needs to be updated in future it can be changed here.
+	 *
+	 * @param string  $message The message to log.
+	 * @param string  $level Info, error, warning, etc.
+	 * @param boolean $exit_on_error For error messages if desired.
+	 * @return void
+	 */
+	private function log( string $message, string $level = 'info', bool $exit_on_error = false ): void {
```
naxoc, line 481: This function should not be necessary anymore. Logging supports multilogging and other tricks now: https://github.com/Automattic/newspack-migration-tools/blob/trunk/docs/logging.md

naxoc (changes requested):
Added a couple of requests, but this should be good to go once we test that the code still works with the recent changes in the NMT.

tests/Command/TestGhostCMSMigrator.php
ronchambers: Removed.  If/when we get the next GhostCMS migration we can/should discuss breaking the command into smaller test-able chunks...

src/Logic/GhostCMSHelper.php
ronchambers: I left the function...but updated all the logging based on the link you shared above, and the "patch" file you had sent me back in November.

naxoc (approved):
Yes! Let's merge this. Thank you Ron 🙏

PR #24, opened 2024-10-03
Fix: Remove trait when command classes use static methods

ronchambers:
This PR will remove the trait from the command classes that use static methods.  The trait was causing a fatal error when run in a WP_CLI context.

Test command:  `wp newspack-migration-tools attachments-get-ids-by-years`

Error:  `PHP Fatal error:  Uncaught Error: Call to private __construct() from scope WP_CLI\Dispatcher\CommandFactory in phar://.../bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Dispatcher/CommandFactory.php:99`

The conflicting line of code: (line 99): `$callable[0] = is_object( $callable[0] ) ? $callable[0] : new $callable[0]();`

It seems that WP_CLI will try to instantiate a new instance when the trait exists on a command class that uses static methods.

The trait has been removed, and doc blocks updated.

If you'd like to test, please follow the "Development" section in the README.md to create a testing plugin, then run the "test command" above again.

---

Tested with WP_CLI version: `WP-CLI 2.11.0`

### DIFF
--- a/src/Command/WpCliCommandTrait.php
+++ b/src/Command/WpCliCommandTrait.php
@@ -8,8 +8,7 @@ use ErrorException;
 /**
  * Utility trait for ensuring singleton instances of WP CLI commands.
  *
- * If your command class uses static functions for commands – don't use this trait as it will cause a fatal error
- * in WP_CLI because WP_CLI will try to call the private __construct() method before calling your static command.
+ * If your command class uses static functions for commands – you shouldn't use this trait.
  */
 trait WpCliCommandTrait {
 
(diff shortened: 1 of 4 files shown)


src/Command/WpCliCommandTrait.php
```diff
@@ -8,7 +8,8 @@
 /**
  * Utility trait for ensuring singleton instances of WP CLI commands.
  *
- * If your command class uses static functions for commands – you probably don't need this trait.
+ * If your command class uses static functions for commands – don't use this trait as it will cause a fatal error
```
naxoc: I think _If your command class uses static functions for commands – you shouldn't use this trait._

Is better here

naxoc (approved):
Nice catch! I forgot to test commands in this repo when I did the large refactoring.

This looks good to go – I put a small change request on the comment in the trait. Feel free to merge.

PR #25, opened 2024-10-04
Remove unneeded inclusion and instantiation for CAP

naxoc:
We now have the closure ensuring we don't instantiate a bunch of things (see #12). We also instantiate a lot later in the bootstrap process now, so we can clean out some of the hoops we had to jump through before. The CAP helper had to initiate the co-authors-plus plugin for us before because we instantiated it immediately and not on a later action than the co-authors-plus plugin itself. 

This PR removes the inclusion of files and instantiations in the CAP helper that the `init` action now has taken care of for us by the time we run our code. 

## How to test
* Run a command that uses the CAP helper


iuravic (approved):
Nice 👍

PR #28, opened 2024-10-14
Move GutenbergBlockManipulator, transformer and transformer command

naxoc:
Move over and minimally refactor to use the new classes. History is move over too using the below. The `git format-patch` is fired in the NCMM and the `git am` command is fired in the NMT.

```bash
git format-patch --stdout --root -- src/Logic/GutenbergBlockManipulator.php > patches/GutenbergBlockManipulator.patch
git am < ../../patches/GutenbergBlockManipulator.patch

git format-patch --stdout --root -- src/Logic/GutenbergBlockTransformer.php > patches/GutenbergBlockTransformer.patch
git am < ../../patches/GutenbergBlockTransformer.patch

git format-patch --stdout --root -- src/Command/General/BlockTransformerCommand.php > patches/BlockTransformerCommand.patch
git am < ../../patches/BlockTransformerCommand.patch
```


.github/workflows/unittest.yml
```diff
@@ -4,7 +4,7 @@ on: [ pull_request ]
 
 jobs:
   unit-tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
naxoc: See https://github.com/apache/beam/issues/32766 – this is temporary. I'll switch it back when it's fixed.

eddiesshop (approved):

.github/workflows/unittest.yml
naxoc: It's back to working – I'll remove the change

PR #36, opened 2024-11-05
Add a user creation helper class

naxoc:
This adds a new class to help create users consistently.

### DIFF
--- a/src/Logic/UsersHelper.php
+++ b/src/Logic/UsersHelper.php
@@ -1,0 +1,233 @@
+<?php
+
+namespace Newspack\MigrationTools\Logic;
+
+use Exception;
+use InvalidArgumentException;
+use Newspack\MigrationTools\Util\Log\CliLog;
+use WP_User;
+
+class UsersHelper {
+
+	/**
+	 * Get a user by an array that contains either the ID, user_email, user_login or user_nicename.
+	 *
+	 * @param array $data Array with one (or more) of the following keys: 'ID', 'user_email', 'user_login', 'user_nicename'.
+	 *
+	 * @return WP_User|false The user if found, false otherwise.
+	 */
+	public static function get_user( array $data ): WP_User|false {
+		$wp_user = false;
+		if ( ! empty( $data['ID'] ) ) {
+			$wp_user = get_user_by( 'ID', $data['ID'] );
+		} elseif ( ! empty( $data['user_email'] ) ) {
+			$wp_user = get_user_by( 'email', $data['user_email'] );
+		} elseif ( ! empty( $data['user_login'] ) ) {
+			$wp_user = get_user_by( 'login', $data['user_login'] );
+		} elseif ( ! empty( $data['user_nicename'] ) ) {
+			$wp_user = get_user_by( 'slug', $data['user_nicename'] );
+		}
+
+		return $wp_user ?? false;
+	}
+
+	/**
+	 * Get a username that is not in use from a desired username.
+	 *
+	 * If the desired username is in use, a counter will be appended to it until an unused username is found.
+	 *
+	 * @param string $desired_username Desired username.
+	 *
+	 * @return string An unused username.
+	 */
+	public static function get_unused_username( string $desired_username ): string {
+		$i = 0;
+		while ( username_exists( $desired_username ) ) {
+			$desired_username .= ( ++$i );
+		}
+		if ( $i > 0 ) {
+			CliLog::get_logger( 'UsersHelper' )->debug( sprintf( 'Generated username: %s', $desired_username ) );
+		}
+
+		return $desired_username;
+	}
+
+	/**
+	 * Get a nicename that is not in use from a desired nicename.
+	 *
+	 * If the desired nicename is in use, a counter will be appended to it until an unused nicename is found.
+	 *
+	 * @param string $desired_nicename Desired nicename.
+	 *
+	 * @return string An unused nicename.
+	 */
+	public static function get_unused_nicename( string $desired_nicename ): string {
+		$i = 0;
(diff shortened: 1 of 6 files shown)


src/Logic/UsersHelper.php
```diff
@@ -0,0 +1,233 @@
...
+class UsersHelper {
+
+	/**
+	 * Get a user by an array that contains either the ID, user_email, user_login or user_nicename.
+	 *
+	 * @param array $data Array with one (or more) of the following keys: 'ID', 'user_email', 'user_login', 'user_nicename'.
+	 *
+	 * @return WP_User|false The user if found, false otherwise.
+	 */
+	public static function get_user( array $data ): WP_User|false {
```
jorostoyanov: I have a concern with the `: WP_User|false` return type.

According to [PHP Docs](https://www.php.net/releases/8.2/en.php#null_false_true_types), this is a 8.2 feature. We currently have a minimum support for 8.1.

We should either bump the PHP requirement to 8.2 or just use `WP_User|bool` to support 8.1

jorostoyanov (changes requested):

src/Logic/UsersHelper.php
naxoc: Good call! Will update :)

src/Logic/UsersHelper.php
```diff
@@ -0,0 +1,243 @@
...
+	 * If the desired username is in use, a counter will be appended to it until an unused username is found.
+	 *
+	 * @param string $desired_username Desired username.
+	 *
+	 * @return string An unused username.
+	 */
+	public static function get_unused_username( string $desired_username ): string {
+		$i = 0;
+		while ( username_exists( $desired_username ) ) {
+			$desired_username .= ( ++$i );
```
jorostoyanov: One thought I have here..

The `user_login` column on WordPress is limited to 60 characters. Does it make sense to limit the concatenation of additional suffix to this?

src/Logic/UsersHelper.php
```diff
@@ -0,0 +1,243 @@
...
+	public static function get_unused_nicename( string $desired_nicename ): string {
+		$i = 0;
+		while ( self::nicename_exists( $desired_nicename ) ) {
+			$desired_nicename .= ( ++$i );
+		}
+		if ( $i > 0 ) {
+			CliLog::get_logger( 'UsersHelper' )->debug( sprintf( 'Generated nicename: %s', $desired_nicename ) );
+		}
+
+		return mb_substr( $desired_nicename, 0, 50 );
```
jorostoyanov: I see that you make use of the limit here and that's great!

src/Logic/UsersHelper.php
```diff
@@ -0,0 +1,243 @@
...
+
+		// If we don't have an email, we'll create an ugly unusable one so that we can create the user.
+		if ( empty( $user_email ) ) {
+			$email_domain = apply_filters( 'nmt_user_email_default_domain', 'example.com' );
+			$user_email   = self::get_short_sha_from_array( $data ) . '@' . $email_domain;
+		}
+
+		// We insist on setting a nicename to avoid WP setting it to the email sans @.
+		if ( empty( $user_nicename ) ) {
+			$user_nicename = ( $data['first_name'] ?? '' ) . ' ' . ( $data['last_name'] ?? '' );
```
jorostoyanov: Can we change this to like:

```php
$user_nicename = implode( ' ', array_filter( [ $data['first_name'], $data['last_name'] ] ) );
```

Just in case we have a last name without the first name. This way, we won't end up with a space at the beginning of the name.

jorostoyanov (changes requested):

src/Logic/UsersHelper.php
naxoc: Absolutely. That is a really good point! That triggered some refactoring to make sure we honor the lengths. I also added tests for them.

src/Logic/UsersHelper.php
naxoc: I ended up changing the logic a bit to make absolutely sure we don't cut off incremented values.

src/Logic/UsersHelper.php
```diff
@@ -0,0 +1,297 @@
...
+	 *
+	 * @param string $desired_nicename Desired nicename.
+	 *
+	 * @return string An unused nicename.
+	 */
+	public static function get_unused_nicename( string $desired_nicename ): string {
+		$original_nicename = $desired_nicename;
+		$desired_nicename  = trim( $desired_nicename );
+		$max_length        = 50;
+		if ( strlen( $desired_nicename ) >= 50 ) {
```
jorostoyanov: Love the changes! Just change `50` in the condition to `$max_length` so it becomes:

```php
if ( strlen( $desired_nicename ) >= $max_length ) {
```

jorostoyanov (changes requested):
Love the new approach! 🚀 

Just a tiny bit of change needed :)

src/Logic/UsersHelper.php
naxoc: Good call. I'm trimming it now 👍

src/Logic/UsersHelper.php
naxoc: Good catch!

naxoc:
@jorostoyanov sorry to push so many changes late in the review process.

Here is what I added:
* A filter for before the user is created.
* A bunch of documentation in the`docs/` dir.
* The "unique identifier" that is saved on users when created.
* A tiny util class to get user ID from usermeta by key and value.
* More tests.

jorostoyanov (approved):
LGTM 🚀 

I committed two changes with typos. Nothing to worry about 🙌 

Thank you! This is fabulous!

PR #38, opened 2024-11-12
Fix log level parsing

naxoc:
I had added an error in how the NMT_LOG_LEVEL constant was checked – this fixes that.


naxoc:
This is tiny. I"ll merge with no review 🤠

PR #44, opened 2024-12-10
fix: use of `strtok` was stripping url of youtube video code.

eddiesshop:
This led to blank blocks with no definitive URL. Just `https://www.youtube.com/watch`.

To test this:

1. new up the Gutenberg generator class: `$g = new \Newspack\MigrationTools\Logic\GutenbergBlockGenerator();`
2. Run this: `$g->get_youtube('N5HbZd9aqR4');`
3. Notice that the URL is not correctly generated.
4. Checkout this branch.
5. Repeat steps 1 & 2.
6. This time, notice that the URL is correctly set: `https://www.youtube.com/watch?v=N5HbZd9aqR4`

### DIFF
--- a/src/Logic/GutenbergBlockGenerator.php
+++ b/src/Logic/GutenbergBlockGenerator.php
@@ -779,7 +779,16 @@ VIDEO;
 	 */
 	private function youtube_block_from_src( $youtube_video_url ) {
 		// Clean the URL from query params as they break the block.
-		$youtube_video_url = strtok( $youtube_video_url, '?' );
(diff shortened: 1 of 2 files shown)


jorostoyanov (changes requested):
Hey @eddiesshop 👋 

It's working as expected! Can we just extend the test suite with the following example URLs to make sure we have it covered:

* https://www.youtube.com/watch?v=N5HbZd9aqR4
* https://www.youtube.com/embed/N5HbZd9aqR4

Thanks! 🙇

PR #47, opened 2024-12-20
feat(GutenbergBlockGenerator): add details block

kariae:
Adds [Details Block](https://wordpress.org/documentation/article/details-block/) generator.

It can be used like the below:

Check the [test file](https://github.com/Automattic/newspack-migration-tools/compare/feat/add-gutenberg-details-block?expand=1#diff-dc318c0766ec8c595b8b3ae7e2d8bbb356ec60b3159f9d2d14e55f9d9c555decR194-R217) to see how to use it.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -1,39 +1,39 @@
 name: Run Unit Tests
 
-on: [ pull_request ]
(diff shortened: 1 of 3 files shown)


eddiesshop (approved):
Works as advertised!

PR #51, opened 2025-01-17
Add: Drupal Migrator and Helper

ronchambers:
docs/fg-helper.md

- Github provides a "rich diff" of markdown files on the "changes" tab or you click these links to compare in browser: [current doc](https://github.com/Automattic/newspack-migration-tools/blob/trunk/docs/fg-helper.md) to [new doc](https://github.com/Automattic/newspack-migration-tools/blob/6e3710720989d76c76015902f944fc2b87e5ca8a/docs/fg-helper.md).

src/Util/FgHelper.php

- Allow blank prefixes for database tables.
- Added check for active premium plugin.
- Added extra filter to fix WordPress not filtering on options if no db row exists.
- If the option row didn't exist, convert "false" into an array.
- Slight tweak of config value usage.


docs/fg-helper.md
```diff
@@ -1,26 +1,123 @@
...
+    [1] horizon.png
+    [2] things.jpg
+    ... etc
+```
+- `fid` File id. DB table `file_managed` is the "canonical" place files live in the DB.
+- `entity_id` Often a node, but can also be a managed file or other things.
+- Fields (applies to Drupal 8 and up) If you have the `config` folder (it's in the code download in backups), you can get a list of all fields on a node type by going to the `config` folder. Let's say you want to see all fields on a node type `book_review` then list files like `ls field.field.node.book_review.*`.  These will be imported into `postmeta`.
+- Images: If you have the `config` folder, look at `field_image` for book_review: `cat field.field.node.book_review.field_image.yml`.
+
+#### Helpful Drupal sql
```
naxoc, line 109: Maybe add a comment here that the below may or may not work depending on the Drupal version.

naxoc (approved):
This looks good. I just added a small comment. :)

PR #54, opened 2025-02-03
Envira gallery migrator

iuravic:
## Change in this PR

- `src/Command/EnviraGalleryMigrator.php`

Added new reusable command `wp newspack-migration-tools envira-gallery-lite-to-gutenberg-gallery` which converts Envira galleries to tiled galleries (tiled gals layout is the most similar to Envira's layout).

Uses the existing shortcode replacement command, provides Envira specific replacement callback.

## Other small changes

- `src/Command/ShortcodesMigrator.php` just exposed post-type selection via argument
- `src/Logic/GutenbergBlockGenerator.php` added missing `<a>` tag surrounding images which use `linkTo` block parameter
- `src/Logic/GutenbergBlockGenerator.php` small helper method added to decode content which was jumbled and too much serialized


kariae (approved):
LGTM

PR #59, opened 2025-02-25
Feat/consolidate getting uinque user login

eddiesshop:
First piece of a concerted effort to consolidate our logic around WP User retrieval and creation.

### DIFF
--- a/src/Logic/UsersHelper.php
+++ b/src/Logic/UsersHelper.php
@@ -106,6 +106,8 @@ class UsersHelper {
 	 * @param string $desired_username Desired username (`wp_users`.`user_login`).
 	 *
 	 * @return string An unused username.
+	 *
+	 * @throws InvalidArgumentException If the desired username is empty.
 	 */
 	public static function get_unused_username( string $desired_username ): string {
 		if ( empty( $desired_username ) ) {
--- a/src/Logic/UsersHelper.php
+++ b/src/Logic/UsersHelper.php
@@ -82,6 +82,11 @@ class UsersHelper {
 
 		$original_user_login = $desired_user_login;
 		$desired_user_login  = sanitize_user( $desired_user_login );
+		
+		if ( is_email( $desired_user_login ) ) {
+			$desired_user_login = substr( $desired_user_login, 0, strpos( $desired_user_login, '@' ) );
+		}
+
 		$max_length          = 60;
 		if ( strlen( $desired_user_login ) >= $max_length ) {
 			$desired_user_login = trim( mb_substr( $desired_user_login, 0, $max_length ) );
--- a/src/Logic/UsersHelper.php
+++ b/src/Logic/UsersHelper.php
@@ -16,6 +16,8 @@ use WP_User;
  */
 class UsersHelper {
 
+	const MAX_USER_LOGIN_LENGTH = 60;
+
 	/**
 	 * Meta key for the unique identifier for users.
 	 */
@@ -82,18 +84,17 @@ class UsersHelper {
 
 		$original_user_login = $desired_user_login;
 		$desired_user_login  = sanitize_user( $desired_user_login );
-		
+
 		if ( is_email( $desired_user_login ) ) {
 			$desired_user_login = substr( $desired_user_login, 0, strpos( $desired_user_login, '@' ) );
 		}
 
-		$max_length          = 60;
-		if ( strlen( $desired_user_login ) >= $max_length ) {
-			$desired_user_login = trim( mb_substr( $desired_user_login, 0, $max_length ) );
+		if ( strlen( $desired_user_login ) >= self::MAX_USER_LOGIN_LENGTH ) {
+			$desired_user_login = trim( mb_substr( $desired_user_login, 0, self::MAX_USER_LOGIN_LENGTH ) );
 			FileLog::get_logger( 'UsersHelper' )->warning(
 				sprintf(
 					'Shortened user login to under %d chars from "%s" to "%s".',
-					$max_length,
+					self::MAX_USER_LOGIN_LENGTH,
 					$original_user_login,
 					$desired_user_login
 				)
@@ -102,7 +103,7 @@ class UsersHelper {
 
 		$i = 0;
 		while ( username_exists( $desired_user_login ) ) {
-			$desired_user_login = self::append_number_and_ensure_length( $desired_user_login, ++$i, $max_length );
+			$desired_user_login = self::append_number_and_ensure_length( $desired_user_login, ++$i, self::MAX_USER_LOGIN_LENGTH );
 		}
 		if ( $i > 0 ) {
 			CliLog::get_logger( 'UsersHelper' )->debug( sprintf( 'Generated user login: %s', $desired_user_login ) );
--- a/src/Logic/UsersHelper.php
+++ b/src/Logic/UsersHelper.php
@@ -81,7 +81,7 @@ class UsersHelper {
 		}
 
 		$original_user_login = $desired_user_login;
-		$desired_user_login  = trim( $desired_user_login );
+		$desired_user_login  = sanitize_user( $desired_user_login );
 		$max_length          = 60;
 		if ( strlen( $desired_user_login ) >= $max_length ) {
 			$desired_user_login = trim( mb_substr( $desired_user_login, 0, $max_length ) );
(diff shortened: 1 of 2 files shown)


src/Logic/UsersHelper.php
```diff
@@ -67,39 +69,87 @@ public static function get_user( array $data ): WP_User|bool {
...
+	 * @param string $user_login The user login to check.
+	 * @param int    $exclude_user_id A user ID to exclude from the check.
 	 *
-	 * @param string $desired_username Desired username.
+	 * @return bool
+	 */
+	public static function is_user_login_unused( string $user_login, int $exclude_user_id = 0 ): bool {
+		global $wpdb;
+
+		$prepared_sql = $wpdb->prepare( "SELECT ID FROM $wpdb->users WHERE user_login = %s", $user_login );
```
naxoc: Why the raw SQL here? `get_user_by( 'login', $user_login)` would be cleaner

src/Logic/UsersHelper.php
```diff
@@ -67,39 +69,87 @@ public static function get_user( array $data ): WP_User|bool {
...
+
+		if ( $exclude_user_id ) {
+			// phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- Reason: $sql_prepared is a prepared statement.
+			$prepared_sql = $wpdb->prepare( "$prepared_sql AND ID <> %d", $exclude_user_id );
+		}
+
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.PreparedSQL.NotPrepared
+		$test = null === $wpdb->get_var( $prepared_sql );
+
+		if ( $test && apply_filters( 'newspack_migration_tools_confirm_user_login_unique_across_coauthors_plus_data', false ) ) {
```
naxoc: Could we call this filter something more generic? There could be other plugins that could filter too. The naming here still ties this to CAP.

I think I'm also confused why the action is called here?

src/Logic/UsersHelper.php
```diff
@@ -67,39 +69,87 @@ public static function get_user( array $data ): WP_User|bool {
...
 	 *
 	 * @return string An unused username.
 	 */
-	public static function get_unused_username( string $desired_username ): string {
-		$original_username = $desired_username;
-		$desired_username  = trim( $desired_username );
-		$max_length        = 60;
-		if ( strlen( $desired_username ) >= $max_length ) {
-			$desired_username = trim( mb_substr( $desired_username, 0, $max_length ) );
+	public static function get_unused_user_login( string $desired_user_login ): string {
```
naxoc: I think I liked the naming with username better. I think only the DB calls it user_login. Keeping username matches it with all other functions having to do with that in WP: https://developer.wordpress.org/?s=username Using 'login' is more of a verb as I see it in WP code.

src/Logic/UsersHelper.php
eddiesshop: Main reason's are:
1. to search exactly for the value provided
2. to make sure we search directly in the DB each time

[`get_user_by()`](https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-user.php#L239-L240) ends up using `sanitize_user()` and a cache for return values.

src/Logic/UsersHelper.php
eddiesshop: 🙏 

Updated this to `nmt_additional_unused_username_check` and made it so that anyone needed to perform additional checks, can tie into it here. I was thinking about it the wrong way before this commit (by bringing in CAP related logic into this).

Whenever there are sites where CAP is still being used, we need to use this filter hook to make sure the desired  `wp_user`.`user_login` is unique even across CAP data

src/Logic/UsersHelper.php
eddiesshop: Cool no worries, I updated the code to go back to using `username`. 

However, I also added some extra documentation and messaging to make sure that people know they are ultimately dealing with the `wp_users`.`user_login` DB field. I've seen a lot of code which gets this mixed up, and I originally wanted to try to clarify this fact.

src/Logic/UsersHelper.php
```diff
@@ -67,36 +69,78 @@ public static function get_user( array $data ): WP_User|bool {
...
+
+		if ( $exclude_user_id ) {
+			// phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- Reason: $sql_prepared is a prepared statement.
+			$prepared_sql = $wpdb->prepare( "$prepared_sql AND ID <> %d", $exclude_user_id );
+		}
+
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.PreparedSQL.NotPrepared
+		$found_no_match = null === $wpdb->get_var( $prepared_sql );
+
+		if ( $found_no_match && has_filter( 'nmt_additional_unused_username_check' ) ) {
```
naxoc: There shouldn't be any reason to check if the filter exists. `apply_filters` does that already. The filtered value needs to be passed to the filter though: `apply_filters( 'nmt_additional_unused_username_check', $found_no_match, $username, $exclude_user_id );` (maybe rename the var to $exists). 

Also – if we can add a docblock over the filter apply it's easier for people to see the intention.

src/Logic/UsersHelper.php
```diff
@@ -67,36 +69,78 @@ public static function get_user( array $data ): WP_User|bool {
...
-		if ( strlen( $desired_username ) >= $max_length ) {
-			$desired_username = trim( mb_substr( $desired_username, 0, $max_length ) );
+		if ( empty( $desired_username ) ) {
+			throw new InvalidArgumentException( 'Desired username (`wp_users`.`user_login`) cannot be empty.' );
+		}
+
+		$original_user_login = $desired_username;
+
+		if ( is_email( $desired_username ) ) {
+			$desired_username = substr( $desired_username, 0, strpos( $desired_username, '@' ) );
```
naxoc: I think using `$desired_username = trim( mb_substr( $desired_username, 0, $max_length ) );` is better here. 

And maybe we should check if it's empty after the substr transformation?

naxoc (changes requested):

src/Logic/UsersHelper.php
```diff
@@ -67,36 +69,88 @@ public static function get_user( array $data ): WP_User|bool {
...
+
+		if ( $username_is_unused ) {
+			/*
+			 * We've confirmed that the `$username` is unused (in the `wp_users` table), but certain plugins might want
+			 * to add their own logic to check for additional conditions. Specifically, if the co-authors-plus
+			 * plugin is installed/activated, there should be additional checks for `$username` uniqueness
+			 * on `wp_terms`.`name` and (`wp_postmeta`.`meta_key`, `wp_postmeta`.`meta_value`) =
+			 * ( 'cap-user_login', $username ) for example.
+			 */
+			$username_is_unused = apply_filters( 'nmt_additional_unused_username_check', $username, $exclude_user_id );
```
naxoc: The arg after the filter name in  `apply_filters()` should always be the value potentially modified. It's a WP convention.

We can also avoid the conditional here by just using:
`return apply_filters( 'nmt_additional_unused_username_check', $username_is_unused, $username, $exclude_user_id );`

naxoc (commented):
Thanks for the changes!

I added a comment about the args on apply_filter.

And the documentation above it is good, but if we can make it a phpdoc block (sorry – I was unclear about that in the comment initially) it would be :chefs_kiss:

src/Logic/UsersHelper.php
eddiesshop: Good point. Done!

eddiesshop: I remember why I had it this way. Say `$username_is_unused = false` at this point, meaning we know for a fact that there is at least another duplicate value in `wp_users`.`user_login` column. Do we want to allow some other plugin/integration to override that assessment and return a `false` value? That's why I only apply the filter if we already know at a basic level it's unique across the `wp_users` table. Does this make sense to keep this way?

naxoc: Hm. Yeah, I see that point too. Let's keep the conditional :) 

But the filter should still pass the filter value I think. Just to not confuse implementers.

naxoc (approved):
This is getting a 👍 from here!

PR #61, opened 2025-03-12
Fix: Attachments Sideload Error: return if error

ronchambers:
## Changes

Since `is_wp_error( $att_id )` is `true`, we should do a `return` so that the function doesn't keep running.  The function already returns `WP_Error` in other places, so returning `WP_Error` here is not a problem.

## History

Looking at the history this bug has been here for a long time.  I assume it never happens because the only way I could cause it was to run the code in "debugger" and stop right before `media_handle_sideload` and delete the `/tmp/` file so that sideload would fail.

## How to test:

Run in "debugger" and stop right before `media_handle_sideload` and delete the `/tmp/` file so that sideload will fail.

### DIFF
--- a/src/Logic/Attachments.php
+++ b/src/Logic/Attachments.php
@@ -116,9 +116,8 @@ class Attachments {
 		if ( is_wp_error( $att_id ) ) {
 			// phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
 			@unlink( $file_array['tmp_name'] );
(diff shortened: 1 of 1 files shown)


ronchambers:
@kariae  or @eddiesshop  - this is an old bug I stumbled upon.  It looks like it came over from NCCM.  It's been tested if one of you can approve.  Thanks!

kariae (approved):
Nice catch!

PR #62, opened 2025-04-07
feat: initial commit of cmd to convert XMLs to JSON

eddiesshop:
The purpose of this PR is to create a mechanism for converting XMLS (and eventually CSVs) to JSON. This is so that we can support our transition to a JSON first migration methodology.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,8 @@
         "php": ">=8.1",
         "monolog/monolog": "^3.7",
         "bramus/monolog-colored-line-formatter": "^3.1",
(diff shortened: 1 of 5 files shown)


eddiesshop:
Team feedback was that this PR isn't really necessary since we have tools like `xq` which can convert an XML to JSON from command line. This same tool could be used to convert a whole directory worth of XML files to JSON.

It's likely we may not have a need for this class at all, so for now I'm closing the PR, but leaving the branch up for a little while longer until we feel absolutely sure we can prune it.

PR #66, opened 2025-05-09
Feat: Allow skip of binary image matching when sideloading

ronchambers:
This PR allows "skipping" of the `maybe_get_existing_attachment_id()` function when importing external files.

Sometimes we want to have two distinct attachments created even if the filename and binary data is the same.  Here is an example:

We're migrating a publisher website that has images at these two urls:

```
example.com/article-one/image.jpg
example.com/article-two/image.jpg
```

These two images have the same filename, and their binary data is the same, so only importing one attachment makes sense, unless the publisher has different alt/caption for each image.  We need two distinct attachments in order to have distinct alt/caption, etc.

Additionally, during migrations we save "old-slug" info into the attachment `post_meta` rows.  If we combine the two images into one, then we could lose our "source of truth" old slug info since only one "old-slug" would be saved into postmeta.  (Yes, we could save "old-slug" as a post meta array with all sources of truths that point to this one attachment, but this can become cumbersome, and still doesn't solve the multiple alt/caption issue).

### DIFF
--- a/src/Logic/Attachments.php
+++ b/src/Logic/Attachments.php
@@ -44,12 +44,16 @@ class Attachments {
 	 * @param array  $args        Optional. Attachment creation argument to override used by the \media_handle_sideload(), used
 	 *                            internally by the \wp_insert_attachment(), and even more internally by the \wp_insert_post().
 	 * @param string $desired_filename Optional. If the file you are importing has a different (or no) file extension than the one
-	 * you want the resulting attachment to have, you can specify it here. Make sure that it
-	 * actually matches the file mime type.
+	 *                                 you want the resulting attachment to have, you can specify it here. Make sure that it
+	 *                                 actually matches the file mime type.
+	 * @param bool   $try_existing Optional. Default true. Try to match an existing file using self::maybe_get_existing_attachment_id().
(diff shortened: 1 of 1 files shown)


ronchambers:
@kariae - this is the solution to the problem we discussed today.  Please review when you can.

(FYI @eddiesshop @iuravic @jorostoyanov - see this PR if you need to force distinct images when doing a migration.)

kariae:
This solution will fix the described issue. My concern is that we'll always have to manually check if we have such a case (the same binary image with a different meta). If so, we'll have to set the flag to false. What about checking for the image meta [when we're checking for the file binary](https://github.com/Automattic/newspack-migration-tools/blob/1d8ce878adeb414be4ae4f0017cb46a97dd86945/src/Logic/Attachments.php#L199).

In the case we discussed (we don't have access to the image meta when we upload it), the above solution won't work, and your flag makes more sense. So, I guess we can merge both solutions. What do you think?

ronchambers:
@kariae - I merged this PR since it successfully ran as part of the BridgeMI CLIs.  I added your suggestion as an [issue here](https://github.com/Automattic/newspack-migration-tools/issues/69).  Thanks!

PR #72, opened 2025-05-31
Add CSVWriter Util class

jorostoyanov:
## Summary

Adds a `CsvWriter` utility class for writing CSV files, with built-in support for setting headers, adding rows, and handling file operations safely.

## Purpose

The `CsvWriter` class makes it easy to write data to CSV files with minimal setup. It manages file creation, prevents header duplication, and throws exceptions if any file operation fails.

## Usage Example

```php

use Newspack\MigrationTools\Util\CsvWriter;

try {
    $csv = new CsvWriter( 'output.csv' ); // The file is created in the current working directory
    $csv->set_header( [ 'Name', 'Email', 'Age' ] );
    $csv->put( [ 'Alice', 'alice@example.com', 30 ] );
    $csv->put( [ 'Bob', 'bob@example.com', 28 ] );
    $csv->close();
} catch ( Exception $e ) {
    error_log( $e->getMessage() );
}
```

## Tests

Unit tests are included in `tests/Util/CsvWriterTest.php` to cover file creation, header handling, row writing, and error conditions.

### DIFF
--- a/src/Util/CsvWriter.php
+++ b/src/Util/CsvWriter.php
@@ -26,11 +26,11 @@ class CsvWriter {
 	public function __construct(
 		private string $filename
 	) {
-        // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_fopen
+		// phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_fopen
 		$this->file_pointer = fopen( getcwd() . '/' . $this->filename, 'a+' );
 
 		if ( false === $this->file_pointer ) {
-            // phpcs:ignore WordPress.Security.EscapeOutput.ExceptionNotEscaped
+			// phpcs:ignore WordPress.Security.EscapeOutput.ExceptionNotEscaped
 			throw new Exception( "Could not open file: {$this->filename}" );
 		}
 	}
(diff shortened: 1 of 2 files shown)


jorostoyanov:
Hey @ronchambers, @Automattic/newspack-launch-and-infrastructure 👋 

I would appreciate your giving a quick look at this PR. Once merged, we can utilize this Util class to generate QA CSV files for each migration and extend it as needed.

Thanks! 🙇‍♂️

kariae:
Looks perfect to me. Can we add support for associative arrays?

jorostoyanov:
> Looks perfect to me. Can we add support for associative arrays?

Sure, @kariae ! How do you imagine the API for this?

**1) Always use `array_values` before `fputcsv`, so you can pass either an indexed or an associative array**

```
fputcsv( $this->file_pointer, array_values( $row ) )
```

This way it can be more readable when populating multiple values, and it's up to the Developer to choose what to pass.

**2) Perform additional checks on whether the number of passed array values matches the header columns**

kariae:
What I meant is to have both of these lines be valid.

```php
$csv->put( [ 'Alice', 'alice@example.com', 30 ] );
$csv->put( [ 'name' => 'Bob', 'age' => 28 ] ); // In this case the email is empty.
```

jorostoyanov:
> What I meant is to have both of these lines be valid.
> 
> ```
> $csv->put( [ 'Alice', 'alice@example.com', 30 ] );
> $csv->put( [ 'name' => 'Bob', 'age' => 28 ] ); // In this case the email is empty.
> ```

@kariae There are cases when you don't need a header row (e.g. when generating a Redirection CSV). In that case, is throwing an exception sufficient? In all other cases, empty values will be added for the non-passed keys.

kariae:
Let's do this for a second iteration of the class and keep it simple for now.

kariae (approved):

PR #74, opened 2025-06-04
Improve Attachment Handling and Duplicate Detection

kariae:
## Changes
- Enhanced duplicate attachment detection with improved filename pattern matching
- Added support for detecting scaled image variants (e.g., `-scaled` suffix)
- Improved handling of numeric suffixes in filenames (e.g., `image-1.jpg`)
- Added better logging for broken attachment detection
- Added support for S3-hosted attachments in broken URL detection
- Added helper methods for attachment retrieval and image source handling

## Technical Details
- Implemented a more robust regex pattern for matching attachment filenames that handles:
  - Base filenames with optional numeric suffixes (e.g., `image-1.jpg`)
  - Scaled image variants (e.g., `image-scaled.jpg`, `image-1-scaled.jpg`)
  - Path-based filenames (e.g., `2023/10/image.jpg`)
- Added `get_attachment_image_src()` helper to bypass Jetpack's Photon CDN
- Improved error handling and logging for broken attachment detection
- Added support for S3-hosted attachments in the broken URL checker

#### Normal Files (1-5MB)
1. **Basic Duplicate Detection**
   - Upload an image (e.g., `test-image.jpg`, ~2MB)
   - Try to upload the same image again
   - Expected: Should return the existing attachment ID instead of creating a duplicate
   - Verify in Media Library that only one copy exists

2. **Scaled Image Variants**
   - Upload a large image that triggers WordPress scaling (e.g., `large-image.jpg`, > 2560px)
   - Check that both original and `-scaled` versions are handled correctly
   - Try to upload the same image again
   - Expected: Should detect and return the  existing attachment ID
   - Verify both original and scaled versions are preserved

3. **Numeric Suffix Handling**
   - Upload `test-image-1.jpg` (~2MB)
   - Upload the same file again, this time renamed to `test-image.jpg`
   - Expected: Should return the existing attachment ID instead of creating a duplicate

#### Large Files (5MB+)
1. **Large File Duplicate Detection**
   - Upload a large image (e.g., `large-photo.jpg`, ~8MB)
   - Try to upload the same file again
   - Expected: Should return the existing attachment ID
   - Verify file integrity is maintained

2. **Large Scaled Images**
   - Upload a very large image (e.g., `huge-photo.jpg`, ~10MB)
   - Let WordPress create scaled versions
   - Try to upload the same file again
   - Expected: Should detect and return the existing attachment ID
   - Verify all scaled versions are properly handled

### DIFF
--- a/src/Logic/Attachments.php
+++ b/src/Logic/Attachments.php
@@ -195,7 +195,8 @@ class Attachments {
 			"SELECT post_id
 			FROM {$wpdb->postmeta}
 			WHERE meta_key = '_wp_attached_file'
-			AND meta_value REGEXP %s;",
+			AND (meta_value LIKE %s OR meta_value REGEXP %s);",
+			$like,
 			$regex_pattern
 		);
 
--- a/src/Logic/Attachments.php
+++ b/src/Logic/Attachments.php
@@ -206,6 +206,8 @@ class Attachments {
 		foreach ( $attachment_ids as $attachment_id ) {
 
 			$candidate_path = get_attached_file( $attachment_id );
+			// Remove the "-scaled" suffix from the candidate path to check the original file size.
+			$candidate_path = str_replace( '-scaled', '', $candidate_path );
 			// Check the file sizes first. It's a fast operation and will save us from having to do the md5 check.
 			if ( ! file_exists( $candidate_path ) || ( filesize( $candidate_path ) !== filesize( $filepath ) ) ) {
 				continue;
--- a/src/Logic/Attachments.php
+++ b/src/Logic/Attachments.php
@@ -207,7 +207,7 @@ class Attachments {
 
 			$candidate_path = get_attached_file( $attachment_id );
 			// Remove the "-scaled" suffix from the candidate path to check the original file size.
-			$candidate_path = str_replace( '-scaled', '', $candidate_path );
+			$candidate_path = self::remove_scaled_suffix_from_image_url( $candidate_path ) ?? $candidate_path;
 			// Check the file sizes first. It's a fast operation and will save us from having to do the md5 check.
 			if ( ! file_exists( $candidate_path ) || ( filesize( $candidate_path ) !== filesize( $filepath ) ) ) {
 				continue;
@@ -218,6 +218,45 @@ class Attachments {
 			}
 		}
 
+		return null;
+	}
+
+	/**
+	 * Remove the '-scaled' suffix from the image URL, or return null if the URL is not a scaled image.
+	 * 
+	 * @param string $url  The image URL.
+	 * @return string|null The image URL without the '-scaled' suffix, or null if $url does not contain a scaled image.
+	 */
+	public static function remove_scaled_suffix_from_image_url( string $url ): string|null {
+		
+		// Supported WordPress image extensions.
+		$wp_supported_extensions = [ 'jpg', 'jpeg', 'png', 'gif', 'webp' ];
+	
+		// Extract filename and extension from URL.
+		$parsed_url = wp_parse_url( $url );
+		if ( ! isset( $parsed_url['path'] ) ) {
+			return null;
+		}
+		$path     = $parsed_url['path'];
+		$filename = basename( $path );
+	
+		// Match WP-supported image filenames that end in -scaled.
+		if ( preg_match( '/^(.*)-scaled\.(' . implode( '|', $wp_supported_extensions ) . ')$/i', $filename, $matches ) ) {
+			// $new_filename = {str_before_scaled} + {str_extension}. phpignore: Squiz.PHP.CommentedOutCode.Found.
+			$new_filename = $matches[1] . '.' . $matches[2];
+	
+			// Replace the filename in the path.
+			$new_path = str_replace( $filename, $new_filename, $path );
+	
+			// Rebuild URL with the modified path and original query string
+			$rebuilt_url = ( isset( $parsed_url['scheme'] ) ? "{$parsed_url['scheme']}://" : '' ) .
+							( isset( $parsed_url['host'] ) ? "{$parsed_url['host']}" : '' ) .
+							$new_path .
+							( isset( $parsed_url['query'] ) ? "?{$parsed_url['query']}" : '' );
+	
+			return $rebuilt_url;
+		}
+	
 		return null;
 	}
 
--- a/src/Logic/Attachments.php
+++ b/src/Logic/Attachments.php
@@ -206,8 +206,6 @@ class Attachments {
 		foreach ( $attachment_ids as $attachment_id ) {
 
 			$candidate_path = get_attached_file( $attachment_id );
-			// Remove the "-scaled" suffix from the candidate path to check the original file size.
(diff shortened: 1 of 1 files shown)


src/Logic/Attachments.php
```diff
@@ -164,31 +164,42 @@ public static function maybe_get_existing_attachment_id( string $filepath, strin
...
-		 * Here's the regex explained:
-		 *  - .+ -- anything first
-		 *  - %s -- is for filename before suffix
-		 *  - -[0-9]+ -- dash and one or more numbers
-		 *  - \%s -- is for filename after suffix
+		 * Constructs a regular expression to find attachments that could be duplicates.
+		 * The regex matches filenames with optional numeric and "-scaled" suffixes,
+		 * where the "-scaled" suffix appears after any numeric suffix if both are present.
+		 * For a base filename like "myimage.jpg", it will match:
+		 * - myimage.jpg
```
iuravic, line 171: I believe there's an issue here where re-uploading a huge image will create a duplicate. Here's an example which can be run for testing -- this img URL actually works, and it returns a 30MB file, so just note that importing it could take a couple of seconds:
```
$path = 'https://ivannptest.newspackstaging.com/wp-content/uploads/30MB_size_demo.jpg';
$id = $attachments->import_external_file( $path );
if ( is_wp_error( $id ) ) {
  WP_CLI::error( $id->get_error_message() );
} else {
  WP_CLI::line( sprintf( '%d %s', $id, $path ) );
}
```

A dupe gets created because of this line below:
```
$candidate_path = get_attached_file( $attachment_id );
```
For a huge non-scaled files, `get_attached_file( $id )` will return the path to -scaled file (because WP will literally use -scaled version everywhere instead of the original huge file), and comparing it to non-scaled `$path` will not be a match.

I believe this line should be updated:
- if `$path` points to a non-scaled file, then `$candidate_path` should get the non -scaled path
- else, it can stay the same

iuravic (changes requested):

src/Logic/Attachments.php
kariae: Nice catch, Ivan! I'm pretty sure I had this in my original code. I'm blaming AI for taking it off 🤣 I added the fix!

kariae: And that's a beautiful picture, btw!

iuravic (changes requested):
Howdy @kariae , I have checked the [8d49cf1](https://github.com/Automattic/newspack-migration-tools/pull/74/commits/8d49cf192885fcc7de6a2f08d6da290d31921943) and would like to show a quick upgrade for that -- I've pushed a demo/proposal commit [44438f4](https://github.com/Automattic/newspack-migration-tools/pull/74/commits/44438f4bad31ebc45ecb5b12c40842586c44c672) which I was thinking you could either:
- reject it completely (just revert it)
- modify freely any part of it
- accept, or any other option you prefer. My idea was just to show some more edge cases in code, let you decide next


This kind of an extended function also handles some edge cases:

- URL contains "-scaled" substring in other places too, but they won't all get replaced, just the one before extension in filename:
```
url: `https://eg.com/path-scaled-images/img-scaled.jpg?b=non-scaled-img` -> `https://eg.com/path-scaled-images/img.jpg?b=non-scaled-img`
```

- image filename itself contains multiple "-scaled" substrings "custom-scaled-image-scaled.jpg", but just the last one before extension gets replaced, not both of them:
```
url: https://eg.com/path-scaled-images/custom-scaled-image-scaled.jpg?b=non-scaled-img -> https://eg.com/path-scaled-images/custom-scaled-image.jpg?b=non-scaled-img
```

- Confirming that simple cases also work correctly:
```
url: https://eg.com/path/img-scaled.jpg -> https://eg.com/path/img.jpg
url: https://eg.com/path/img.jpg -> null (it's not a -scaled image URL)
```

- This example contains multiple "-scaled" substrings, but it's still not a scaled image URL:
```
url: https://eg.com/path-scaled-images/img.jpg?b=non-scaled-img -> null (it's not a -scaled image URL)
```

- Just do this for images, not other file types:
```
url: https://eg.com/path/not-an-image-scaled.txt -> null (it's not a -scaled image URL)
```

- Complex string, confirming that regex is OK:
```
url: http://another.net/a/b/c/d/e/f-g-h-i-j-k-l-m-n-o-p-scaled.gif?query=param#fragment -> http://another.net/a/b/c/d/e/f-g-h-i-j-k-l-m-n-o-p.gif?query=param
```

Feel free not to use this at all if you think it's an overkill ;) I just thought I'd show and share a similar but a bit different approach.

P.S. my personally preference is to use regular old non-static methods, but I had to make this one `static` because it's called from the existing static `maybe_get_existing_attachment_id`.

---

**UPDATE**
I originally did this method for URLs, not local paths, but it works equally well on both. I did the URL in mind for a thing I am working on for a Publisher. So if you do decide that you like the idea, perhaps we can rename it from `remove_scaled_suffix_from_image_url` to `remove_scaled_suffix_from_image_path`, and add a comment that it works both on local paths and URL.

Examples:
```
path: /path-scaled-images/img-scaled.jpg -> /path-scaled-images/img.jpg
path: /path-scaled-images/custom-scaled-image-scaled.jpg -> /path-scaled-images/custom-scaled-image.jpg
path: /path/img-scaled.jpg -> /path/img.jpg
path: /path/img-scaled.jpeg -> /path/img.jpeg
path: /img-scaled.png -> /img.png
path: /a/b/c/d/e/f-g-h-i-j-k-l-m-n-o-p-scaled.gif -> /a/b/c/d/e/f-g-h-i-j-k-l-m-n-o-p.gif
path: /path-scaled-images/img.jpg -> null
path: /path/img.jpg -> null
path: /path/not-an-image-scaled.txt -> null
```

iuravic:
I now tested all remaining scenarios I could think of, and am noticing one more edge case -- if image `testtwo.png` existis in the media library, and we try and upload the identical file/image but called `testtwo-1.png`, it will import a new one.

It feels to me personally that this is OK behavior, and no need to optimize too much and that we can leave it as is. But still sharing the finding.

kariae:
Hey @iuravic 👋 I agree that the `-scaled` string can be in other places in the filepath, but I don't think all these edge cases are worth the overkill. I do like regex, and I'm tempted to give this a thumbs up, but I believe that we can catch most of them with the below change, and if we don't see a few it would still be fine to have a couple duplicate files (which I still think it would be rare). We don't have to keep track of image extensions, either.

```diff
-        $candidate_path = str_replace( '-scaled', '', $candidate_path );
+        $candidate_path = str_replace( '-scaled.', '.', $candidate_path );
```

What do you think?

iuravic:
> What do you think?

Sure thing, @kariae , it seems that this will match more than 99% of cases with minimal effort.

I've now reverted my demo changes, so that you don't have to do it ;) and it's now back to where you left it to add the `.`, then we merge 👍

kariae:
Thank you, @iuravic. Please give this a final check, and I'll merge it.

iuravic (approved):
Works great now! 👏 👏 👏

PR #78, opened 2025-06-17
Enhance Guest Contributors Helper Documentation and Functionality

kariae:
# Add Guest Contributor Creation and Assignment Methods

## Overview

This PR adds two new methods to the `GuestContributorsHelper` class to enhance the functionality for working with guest contributors:

1. `create_or_get_contributor()` - A wrapper method for creating or retrieving guest contributor users
2. `assign_contributors_to_post()` - A method for assigning guest contributors to posts using Co-Authors Plus

### New Method: `create_or_get_contributor()`

**Purpose**: Provides a convenient wrapper around `UsersHelper::create_or_get_user()` that automatically sets the role to the guest contributor role.

**Parameters**:
- `$data` (array): User data array compatible with `wp_insert_user()`
- `$unique_identifier` (string): Unique identifier for the user

**Returns**: `WP_User|WP_Error` - User object if created/found, or WP_Error on failure

**Key Features**:
- Validates Newspack plugin is active before proceeding
- Automatically sets the role to `Guest_Contributor_Role::CONTRIBUTOR_NO_EDIT_ROLE_NAME`
- Delegates to `UsersHelper::create_or_get_user()` for the actual user creation/retrieval logic

**Example Usage**:
```php
$user_data = [
    'display_name' => 'John Smith',
    'user_email'   => 'john.smith@example.com',
    'user_login'   => 'johnsmith',
];

$user = GuestContributorsHelper::create_or_get_contributor($user_data, 'john-smith-unique');
if (is_wp_error($user)) {
    // Handle error
}
```

### New Method: `assign_contributors_to_post()`

**Purpose**: Assigns guest contributor users to a specific post using the Co-Authors Plus plugin.

**Parameters**:
- `$post_id` (int): The ID of the post to assign contributors to
- `$contributor_ids` (array): Array of contributor user IDs to assign

**Returns**: `true|WP_Error` - True on success, WP_Error on failure

**Key Features**:
- Validates that Co-Authors Plus plugin is active
- Uses the global `$coauthors_plus` object to assign authors
- Provides detailed error messages for debugging
- Uses the 'id' parameter for direct user ID assignment

**Example Usage**:
```php
$post_id = 123;
$contributor_ids = [1, 2, 3];

$result = GuestContributorsHelper::assign_contributors_to_post($post_id, $contributor_ids);
if (is_wp_error($result)) {
    // Handle error
}
```

## Dependencies

- Newspack Plugin >= 6.2.0 (for guest contributor role functionality)
[...]

PR #80, opened 2025-06-18
Feat: Guest Contributors Enhancements: Ron's Review

ronchambers:
Ron's review of https://github.com/Automattic/newspack-migration-tools/pull/78

I found 3 issues:

1) The rare possibility of an Exception wasn't being caught.
2) The UsersHelper's exceptions don't have codes, so WP_Error wasn't being created properly.  I added a bunch of comments on why and how to fix...but in the end we're probably just better setting the code to `ERROR_USERS_HELPER`.
3) The Newspack plugin exists within PHPUNIT [here](https://github.com/Automattic/newspack-migration-tools/blob/9303059c8b9ed1b7c72548c74fb1732157bc5c13/tests/bootstrap.php#L40) so it's ok to just use the role const from that plugin in the tests.

### DIFF
--- a/src/Logic/GuestContributorsHelper.php
+++ b/src/Logic/GuestContributorsHelper.php
@@ -37,8 +37,11 @@ class GuestContributorsHelper {
 		$data['role'] = Guest_Contributor_Role::CONTRIBUTOR_NO_EDIT_ROLE_NAME;
 		try {
 			return UsersHelper::create_or_get_user( $data, $unique_identifier );
-		} catch ( \InvalidArgumentException $e ) {
-			return new WP_Error( $e->getCode(), $e->getMessage() );
+		} catch ( \Exception $e ) {
+			// WP_Error() requires the first argument "code" to be not empty (which also means not "0").
+			// Attempt to retrieve code from exception, else default the code to the exception type.
+			// Don't use the null coalescing operator "??", as it will allow the value of 0 to be accepted.
+			return new WP_Error( $e->getCode() ? $e->getCode() : get_class( $e ), $e->getMessage() );
 		}
 	}
 
--- a/tests/Logic/GuestContributorsHelperTest.php
+++ b/tests/Logic/GuestContributorsHelperTest.php
@@ -2,13 +2,12 @@
 
 namespace Newspack\MigrationTools\Tests\Logic;
 
+use Newspack\Guest_Contributor_Role;
 use Newspack\MigrationTools\Logic\GuestContributorsHelper;
 use WP_Error;
 use WP_UnitTestCase;
 
 class GuestContributorsHelperTest extends WP_UnitTestCase {
-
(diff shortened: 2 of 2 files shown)


src/Logic/GuestContributorsHelper.php
```diff
@@ -37,8 +37,11 @@ public static function create_or_get_contributor( array $data, string $unique_id
 		$data['role'] = Guest_Contributor_Role::CONTRIBUTOR_NO_EDIT_ROLE_NAME;
 		try {
 			return UsersHelper::create_or_get_user( $data, $unique_identifier );
-		} catch ( \InvalidArgumentException $e ) {
-			return new WP_Error( $e->getCode(), $e->getMessage() );
+		} catch ( \Exception $e ) {
```
ronchambers, line 40: `UsersHelper::create_or_get_user()` may return multiple exception types, so just use the more general Exception instead.  (Otherwise `throw new Exception` would have not been caught).

src/Logic/GuestContributorsHelper.php
```diff
@@ -37,8 +37,11 @@ public static function create_or_get_contributor( array $data, string $unique_id
...
+			// WP_Error() requires the first argument "code" to be not empty (which also means not "0").
+			// Attempt to retrieve code from exception, else default the code to the exception type.
+			// Don't use the null coalescing operator "??", as it will allow the value of 0 to be accepted.
+			return new WP_Error( $e->getCode() ? $e->getCode() : get_class( $e ), $e->getMessage() );
```
ronchambers, line 44: Instead of worrying about all that "empty" "0" stuff, we could just hard code to this value:

```suggestion
			return new WP_Error( 'ERROR_USERS_HELPER`, $e->getMessage() );
```

tests/Logic/GuestContributorsHelperTest.php
```diff
@@ -2,14 +2,13 @@
...
 namespace Newspack\MigrationTools\Tests\Logic;
 
+use Newspack\Guest_Contributor_Role;
 use Newspack\MigrationTools\Logic\GuestContributorsHelper;
 use WP_Error;
 use WP_UnitTestCase;
 
 class GuestContributorsHelperTest extends WP_UnitTestCase {
 
-	const NEWSPACK_CONTRIBUTOR_ROLE = 'contributor_no_edit';
```
ronchambers, line 11: The Newspack Plugin exists in PHPUNIT [see here](https://github.com/Automattic/newspack-migration-tools/blob/9303059c8b9ed1b7c72548c74fb1732157bc5c13/tests/bootstrap.php#L40).

So we can just use the role directly from the plugin.

ronchambers:
@kariae if you like the changes, please merge into your branch.   Otherwise let me know your thoughts.

kariae (approved):

Issue #68, opened 2025-05-14
Blocks Generator Enhancements: Paragraph


jorostoyanov:
## The Issue(s)

The `get_paragraph()` method of the `GutenbergBlockGenerator` class needs several updates to match all supported features of the Paragraph Block.

The [current](https://github.com/Automattic/newspack-migration-tools/blob/1d8ce878adeb414be4ae4f0017cb46a97dd86945/src/Logic/GutenbergBlockGenerator.php#L469) implementation has 5 available parameters:
* `string` $paragraph_content
* `string` $anchor
* `string` $text_color
* `string` $font_size
* `array` $additional_css_classes

The attributes below cannot be added currently in any way:
1) `NEW` text align support
2) `BUG` text color support. [Currently existing](https://github.com/Automattic/newspack-migration-tools/blob/1d8ce878adeb414be4ae4f0017cb46a97dd86945/src/Logic/GutenbergBlockGenerator.php#L476C12-L476C20), but with a wrong array key, therefore not applied
3) `NEW` custom attributes
3.1) Drop Cap support
3.2) Direction (LTR | RTL) support
3.3) Letter Spacing
3.4) Dimensions (padding/margin)
3.5) Letter Case
3.6) Decoration
3.7) Letter Spacing
3.8) Appearance (aka Font Weight)

## The Proposal

Personally, I think we don't need to explicitly define every available attribute as a method parameter. Instead, we can have an optional `array` parameter `$attributes` to define everything we need. Then, in the method definition, we can check whether a specific attribute is specified and apply the necessary changes to class names.

The current `$anchor`, `$text_color`, and `$font_size` parameters can be deprecated, but will still be supported for a short period.

## The Implementation

I) In order not to break anything currently using the method, we can just define `$attributes` as the 6th parameter, like so:

```
public function get_paragraph( $paragraph_content, $anchor = '', $text_color = '', $font_size = '', array $additional_css_classes = [], array $attributes = [] ) {
    // And in order to force the usage of this new parameter, to deprecate `$anchor`, `$text_color` and `$font_size`
    if ( ! empty( $anchor ) ) {
         _deprecated_argument( __FUNCTION__, '2.0', '$anchor should be set in the $attributes parameter' );

        $attributes['anchor'] = $anchor;
    }
}
```

II) At some point, we can remove the deprecated parameters and leave the method definition as followed:

```
public function get_paragraph( $content, array $attributes = [], array $additional_css_classes = [] ) {
    // ...
[...]

PR #81, opened 2025-06-22
Get image block for an externally hosted image

iuravic:
### Image block for external images

Adds the `get_external_image()` method which returns the `core/image` block for an externally hosted image:

image

### Standards
- [x] PHPCS
- [x] Tested during real publisher migration

### DIFF
--- a/src/Logic/GutenbergBlockGenerator.php
+++ b/src/Logic/GutenbergBlockGenerator.php
@@ -329,6 +329,65 @@ class GutenbergBlockGenerator {
 	}
 
 	/**
+	 * Generate an Image Block item from an external image URL.
+	 *
+	 * @param string  $image_url              Image URL.
+	 * @param ?string $image_caption          Image caption.
+	 * @param ?string $image_alt              Image alt text.
+	 * @param ?string $size                   Image size, full by default.
+	 * @param ?bool   $link_to_attachment_url Whether to link to the attachment URL or not. Defaults to true.
+	 * @param ?string $classname              Media HTML class.
+	 * @param ?string $align                  Image alignment (left, right).
+	 * @param ?string $custom_link            If provided, will set custom link. Overrides $link_to_attachment_url.
+	 *
+	 * @return array Image block with external image URL, to be used in the serialize_blocks function to get the raw content of a Gutenberg Block.
+	 */
+	public function get_external_image( string $image_url, ?string $image_caption = null, ?string $image_alt = null, ?string $size = 'full', ?bool $link_to_attachment_url = true, ?string $classname = null, ?string $align = null, ?string $custom_link = null ): array {
+		// Validate size.
+		if ( ! in_array( $size, [ 'thumbnail', 'medium', 'large', 'full' ] ) ) {
+			$size = 'full';
+		}
+
+		// Add <a> link to attachment URL.
+		$a_opening_tag = '';
+		$a_closing_tag = '';
+		if ( $custom_link || $link_to_attachment_url ) {
+			if ( $custom_link ) {
+				$link = $custom_link;
+			} elseif ( $link_to_attachment_url ) {
+				$link = $image_url;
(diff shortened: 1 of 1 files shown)


kariae:
Can we refactor the [get_image](https://github.com/Automattic/newspack-migration-tools/blob/ce32eb1e80bdc5d78ec4fba9fd9b53b4bcba2420/src/Logic/GutenbergBlockGenerator.php#L280) method to add the externally hosted image use case? It's the same block (technically with a different attribute). The only difference is that we [don't need an Attachment ID](https://github.com/Automattic/newspack-migration-tools/blob/ce32eb1e80bdc5d78ec4fba9fd9b53b4bcba2420/src/Logic/GutenbergBlockGenerator.php#L292).

I'm asking if we can have one method per Gutenberg block instead of one method per Gutenberg block use case. What do you think?

iuravic:
> Can we refactor the [get_image](https://github.com/Automattic/newspack-migration-tools/blob/ce32eb1e80bdc5d78ec4fba9fd9b53b4bcba2420/src/Logic/GutenbergBlockGenerator.php#L280) method to add the externally hosted image use case? It's the same block (technically with a different attribute). The only difference is that we [don't need an Attachment ID](https://github.com/Automattic/newspack-migration-tools/blob/ce32eb1e80bdc5d78ec4fba9fd9b53b4bcba2420/src/Logic/GutenbergBlockGenerator.php#L292).
> 
> I'm asking if we can have one method per Gutenberg block instead of one method per Gutenberg block use case. What do you think?

@kariae , makes sense. I'll admit, I was in a lot of rush when creating this. I did think about the same thing but was too hasty. Glad that someone (you) helped me during times of stress ;)

I thought the same thing first. Then I noticed that there are these different and same arguments. In my PR we now have two methods:
```
get_image from attachment
- unique args
	$attachment_post
- common args
	$size
	$link_to_attachment_url
	$classname
	$align
	$custom_link

get_external_image
- unique args
	string $image_url
	$image_caption
- common args
	$size
	$link_to_attachment_url
	$classname
	$align
	$custom_link
```

And they now duplicate the image block creation part. What they do differently is prepare arguments for the image block in different ways. To improve it, I'd propose to split into three methods:
```
get_image_block
- this is the main method we already have -- it's now called get_image I'd just rename it to get_image_block
- I would refactor it to take all the possible arguments for the image block (external or from attachment) and return the block

And create these other two small methods which are just wrappers for get_image_block:

get_image_block_from_attachment
- takes arguments specific for attachment, i.e. $attachment_post
=> calls and returns get_image_block

get_external_image_block
takes arguments specific for external image, i.e. $image_url, $image_caption
=> calls and returns get_image_block
```

What do you think?

----

**UPDATE**, I realized we should avoid renaming the current `get_image` because other methods/classes are usng it. Like this then:
- `get_image` -- it exists, it gets image block for attachment, we keep it as it is
  - we just refactor code which gets image block into `get_image_block`
  => it prepares arguments from attachment then calls `get_image_block`
- `get_external_image`
  - we add this new one, a wrapper
  => it prepares arguments from external URL then calls `get_image_block`
- `get_image_block`
  - new one with reusable code

kariae:
If we add the `_block`, we'll need to add it to all the other blocks to keep consistency (not a big deal, but ...).

The needed arguments for these builders can be tricky. For example, in this case, we have some common and some for specific use cases, and that's on me. I didn't think about it first when building this. @jorostoyanov describes a better approach here https://github.com/Automattic/newspack-migration-tools/issues/68. I think that's what we should do in the long run.

But to unblock this, as we don't have capacity for such a refactor, let's merge this as is (creating a second method for the external URLs use case).

kariae (approved):

jorostoyanov:
Thank you @kariae 🙇 

I wanted to give this reference and take part in the discussion, but lost the issue where it has been discussed 😅

PR #83, opened 2025-06-25
Add Simple Local Avatars initialization in constructor

kariae:
Fix the Simple Local Avatar helper class.

## Test This PR

1. Set a user avatar using the `assign_avatar` method.

```php
// 1 is the user ID, 2 is the attachment ID.
$this->simple_local_avatars_logic = new SimpleLocalAvatars();
$this->simple_local_avatars_logic->assign_avatar( 1, 2 );
```

2. Notice the PHP error `Fatal error: Uncaught Error: Call to a member function assign_new_user_avatar() on null`
3. Pull this PR.
4. Re-run the code from step 1.
5. Check that the user got the avatar set in their profile.


jorostoyanov (approved):
LGTM 👌

PR #84, opened 2025-06-25
Add unique identifier support for post's creation

kariae:
This PR introduces support for uniquely identifying posts in the migration tools by leveraging a custom meta key. It provides methods to create or retrieve posts based on a unique identifier, which helps prevent duplication and enables idempotent post creation.

### Changes

- **Added:** `UNIQUE_POST_IDENTIFIER_META_KEY` constant to store the unique identifier meta key for posts.
- **Added:** `create_or_get_post( array $data, string $unique_identifier )`
  - Searches for an existing post using the unique identifier.
  - If found, returns the existing post ID.
  - If not found, creates a new post with the provided data and sets the unique identifier as post meta.
- **Added:** `get_post_by_unique_identifier( string $unique_identifier )`
  - Retrieves a post ID by searching for the unique identifier in post meta.
  - Returns false if not found, and logs an error if the identifier is empty.

### Motivation

- Prevents duplicate posts during migrations or programmatic imports.
- Enables safe, idempotent operations by ensuring posts are only created once for a unique identifier.
- Improves error logging and traceability when dealing with post creation and lookup.

### Testing

- Use `Posts::create_or_get_post()` multiple times with the same unique identifier. Only one post should be created, and subsequent calls should return the same post ID.
- Use `Posts::get_post_by_unique_identifier()` to verify the correct post ID retrieval.
- Confirm that an error is logged if an empty identifier is supplied.

### DIFF
--- a/src/Logic/Posts.php
+++ b/src/Logic/Posts.php
@@ -738,7 +738,7 @@ SQL;
 	 */
 	public static function get_post_by_unique_identifier( string $unique_identifier ): int|false {
 		if ( empty( $unique_identifier ) ) {
-			FileLog::get_logger( 'Posts' )->error(
+			FileLog::get_logger( __CLASS__ )->error(
 				'Value is empty. Refusing to find a post with empty values.',
 				[
 					'unique_identifier' => $unique_identifier,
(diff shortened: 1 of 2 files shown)


jorostoyanov (approved):
Looking good! Thanks, @kariae 🚀

PR #85, opened 2025-06-25
Add filtered_items method to JsonIterator for JSON file item 

kariae:
This PR introduces a new method, `filtered_items`, to the `JsonIterator` class.
The new method allows you to iterate over objects in a JSON file and yield only objects with a specified key matching a specified value.

### New Method

```php
public function filtered_items( string $json_file, string $key, string $value ): iterable
```

- **$json_file:** Path or URL to a JSON file (root should be an array of objects).
- **$key:** The key to filter objects by.
- **$value:** The value must match the specified key.

### Usage Example

Given the following JSON file (save as `sample.json`):

```JSON
[
  { "id": 1, "type": "post", "author": "alice" },
  { "id": 2, "type": "page", "author": "bob" },
  { "id": 3, "type": "post", "author": "carol" },
  { "id": 4, "type": "post", "author": "bob" }
]
```

#### Example 1: Filter by type = "post"

```php
$iterator = new JsonIterator();
foreach ( $iterator->filtered_items( 'sample.json', 'type', 'post' ) as $item ) {
    print_r( $item );
}
```

Expected Output:

```php
stdClass Object
(
    [id] => 1
    [type] => post
    [author] => alice
)
stdClass Object
(
    [id] => 3
    [type] => post
    [author] => carol
)
stdClass Object
(
    [id] => 4
    [type] => post
    [author] => bob
)
```

#### Example 2: Filter by author = "bob"

```php
$iterator = new JsonIterator();
foreach ( $iterator->filtered_items( 'sample.json', 'author', 'bob' ) as $item ) {
    print_r( $item );
}
```

Expected Output:

```php
stdClass Object
(
    [id] => 2
    [type] => page
    [author] => bob
)
stdClass Object
(
    [id] => 4
    [type] => post
    [author] => bob
)
```

### Additional Changes

- Minor improvement: Use `isset` instead of `empty` when checking the result of `jq` in `count_json_array_entries`, because `jq` can return `0` when the JSON array is empty and this breaks the method.

### DIFF
--- a/src/Util/JsonIterator.php
+++ b/src/Util/JsonIterator.php
@@ -117,7 +117,7 @@ class JsonIterator {
 	 *
 	 * @return iterable
 	 */
-	public function filtered_items( string $json_file, string $key, string $value ): iterable {
+	public function filtered_items( string $json_file, string $key, ?string $value = null ): iterable {
 		$file_exists = str_starts_with( $json_file, 'http' ) ? $this->url_responds( $json_file ) : file_exists( $json_file );
 
 		if ( ! $file_exists ) {
@@ -128,7 +128,13 @@ class JsonIterator {
 		try {
 			$items = Items::fromFile( $json_file );
 			foreach ( $items as $item ) {
-				if ( isset( $item->$key ) && $item->$key === $value ) {
+				// If value is null, only check if the key exists.
+				if ( null === $value ) {
+					if ( isset( $item->$key ) ) {
+						yield $item;
+					}
+				} elseif ( isset( $item->$key ) && $item->$key === $value ) {
+					// If value is set, check both key existence and value equality
 					yield $item;
 				}
 			}
--- a/src/Util/JsonIterator.php
+++ b/src/Util/JsonIterator.php
@@ -118,12 +118,25 @@ class JsonIterator {
 	 * @return iterable
 	 */
 	public function filtered_items( string $json_file, string $key, string $value ): iterable {
-		$items = Items::fromFile( $json_file );
-		foreach ( $items as $item ) {
-			if ( isset( $item->$key ) && $item->$key === $value ) {
-				yield $item;
-			}
+		$file_exists = str_starts_with( $json_file, 'http' ) ? $this->url_responds( $json_file ) : file_exists( $json_file );
+
+		if ( ! $file_exists ) {
+			NMT::exit_with_message( sprintf( 'File does not exist: %s', $json_file ), [ $this->file_logger ] );
+			return new \EmptyIterator();
 		}
+
+		try {
+			$items = Items::fromFile( $json_file );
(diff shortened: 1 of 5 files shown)


src/Util/JsonIterator.php
```diff
@@ -108,6 +108,37 @@ public function items( string $json_file, array $options = [] ): iterable {
...
+		}
+
+		try {
+			$items = Items::fromFile( $json_file );
+			foreach ( $items as $item ) {
+				if ( isset( $item->$key ) && $item->$key === $value ) {
+					yield $item;
+				}
+			}
+		} catch ( Exception $o_0 ) {
```
jorostoyanov: Can we change `$o_0` to `$e`, following the NMT codebase? :)

src/Util/JsonIterator.php
```diff
@@ -108,6 +108,37 @@ public function items( string $json_file, array $options = [] ): iterable {
...
+			NMT::exit_with_message( sprintf( 'Could not read the JSON from: %s', $json_file ), [ $this->file_logger ] );
```
jorostoyanov: It would be nice to have the message from the thrown exception as a bonus :)

jorostoyanov (changes requested):
LGTM :)

One minor change and two suggestions. Please, ignore if not applicable.

Does it make sense to change the `$value` parameter to optional? So, we have the ability to match items by just passing the keys, meaning "Get all items that have this key? or Get all Items that have this key with this value?"

Again, ignore if this doesn't make sense for now :)

src/Util/JsonIterator.php
kariae: That's what happens when you accept the changes from Copilot 😅 That's said o_0 is cooler than "e" 🤣 I changed it in both places.

kariae:
> Does it make sense to change the $value parameter to optional? So, we have the ability to match items by just passing the keys, meaning "Get all items that have this key? or Get all Items that have this key with this value?"

I've needed this exact functionality an hour ago, I'll push a commit for it.

kariae:
@jorostoyanov, I added the functionality with unit tests :)

src/Util/JsonIterator.php
naxoc: I think Copilot learned that one from me 🙊 I love using `o_0`

tumblr_inline_p34zg66Ryp1t2lh65_400

jorostoyanov:
@kariae The functionality in this PR got merged with the one from [this PR](https://github.com/Automattic/newspack-migration-tools/pull/84).

Apart from that, everything looks good!

Thanks! 🙇

src/Util/JsonIterator.php
kariae: We should move toward o_0 vs. e, just saying :D

PR #86, opened 2025-07-03
Collections Helper

jorostoyanov:


jorostoyanov:
Hey @ronchambers :wave:

Can I ask you to have a quick look at this PR and validate the API. I already used it on a Project, but I need a second pair of eyes to take a look at the implementation before I write some tests :)

Thanks! 🙇

src/Logic/CollectionsHelper.php
```diff
@@ -0,0 +1,346 @@
...
+
+	public const UNIQUE_COLLECTION_IDENTIFIER_META_KEY          = '_nmt_collection_uniqid';
+	public const UNIQUE_COLLECTION_SECTION_IDENTIFIER_META_KEY  = '_nmt_collection_section_uniqid';
+	public const UNIQUE_COLLECTION_CATEGORY_IDENTIFIER_META_KEY = '_nmt_collection_category_uniqid';
+
+	/**
+	 * Collection metadata map.
+	 */
+	const COLLECTION_META_MAP = [
+		'thumbnail'      => '_thumbnail_id',
```
ronchambers: Very minor but I like "_id" when it's an integer.  So maybe:
```suggestion
		'thumbnail_id'      => '_thumbnail_id',
```

src/Logic/CollectionsHelper.php
```diff
@@ -0,0 +1,346 @@
...
+use Newspack\MigrationTools\Traits\Has_Unique_Identifier;
+use Newspack\Optional_Modules\Collections;
+use WP_Error;
+use WP_Term;
+
+/**
+ * Collections Helper class.
+ */
+class CollectionsHelper {
+	use Has_Unique_Identifier;
```
ronchambers: @jorostoyanov it looks like some of this trait logic might already be in a Helper.

See https://github.com/Automattic/newspack-migration-tools/blob/a64d7b63cfdd22d82160762ef1363eda19452719/src/Logic/Posts.php#L17

And https://github.com/Automattic/newspack-migration-tools/blob/a64d7b63cfdd22d82160762ef1363eda19452719/src/Logic/Posts.php#L739

Also https://github.com/Automattic/newspack-migration-tools/blob/a64d7b63cfdd22d82160762ef1363eda19452719/src/Logic/Taxonomy.php#L22-L27

Maybe those existing functions could be used (or enhanced) instead of the trait?

jorostoyanov: My idea was to update those classes later with the trait in order to encapsulate the whole Unique Identifier logic in a single place. Since the validation logic and returned errors are the same.

The only change is the meta table of the unique identifier and meta key.

ronchambers: Makes sense :-)

iuravic:
@jorostoyanov , kudos for the amazingREADME 🙇

ronchambers:
@jorostoyanov @kariae - since this PR was already run for America Mag (Collections Demo), which looks good from a tech point of view, can we merge this PR into trunk since these are all new files (no conflicts) and America Mag is on a tight timeline?

jorostoyanov:
From my side, I wanted to write some tests, but we can add them later.

And a quick change to apply while I was using the helper for another Publisher.

If you agree, I can merge without tests and write them later 😅

ronchambers (approved):

ronchambers:
Approved, since these are all new files.  Feel free to push the "quick change" you mentioned then squash merge.  Put tests in new PR 😄  Thank you!

ronchambers:
(Tomorrow is cool or Friday Europe time).

src/Logic/CollectionsHelper.php
jorostoyanov: Agreed and changed!

jorostoyanov:
@ronchambers Now merged!

Promise to add tests in a new PR ☺️

ronchambers:
Thanks!  I added two [minor issues](https://github.com/Automattic/newspack-migration-tools/issues) for future...

jorostoyanov:
> Thanks! I added two [minor issues](https://github.com/Automattic/newspack-migration-tools/issues) for future...

@ronchambers Can you update the link to the minor issues, as it currently leads to /issues

ronchambers:
@jorostoyanov 
1. Verify minor tweak to the [docs](https://github.com/Automattic/newspack-migration-tools/issues/97)
2. Error checking needs some [refactoring](https://github.com/Automattic/newspack-migration-tools/issues/98)

jorostoyanov:
@ronchambers 

Thank you!
Those are addressed now here — https://github.com/Automattic/newspack-migration-tools/pull/102

PR #87, opened 2025-07-07
Fix User Email Generation and User Lookup Logic

kariae:
## Overview

This PR addresses two critical issues in the `UsersHelper` class:

1. **Improved email shortening logic** - Fixes email generation to preserve domain parts while ensuring proper length constraints
2. **Enhanced user lookup behavior** - Ensures users are only retrieved by unique identifier, preventing unintended user matching

### Fix 1: Improved Email Shortening Logic (`get_unused_fake_email`)

The previous implementation had a bug that caused long emails to be improperly shortened, potentially resulting in emails that were still too long or had truncated domains. It also appended the numbers prefix to the previously attempted email address.

**Solution**: Refactored the email shortening logic to:
- Preserve the full domain part when shortening emails
- Calculate the maximum local part length as `(96 - domain_length - 1)` to ensure total email stays under 100 characters
- Use `explode('@', $desired_email)` to properly separate local and domain parts
- Maintain email validity throughout the shortening process
- Append the count of iterations to the original email

### Fix 2: Unique Identifier Only User Lookup (`create_or_get_user`)

The method was falling back to searching for users by email, login, or nicename when the unique identifier wasn't found, which could lead to unintended user matching and data inconsistencies, which we were handling by generating fake unused email addresses, usernames, and nicenames.

**Solution**: Modified the user lookup logic to:
- Only search for users by their unique identifier
- Remove fallback searches by other fields (email, login, nicename)
- Ensure that users with the same email/login but different unique identifiers create separate user accounts
- Maintain data integrity by preventing accidental user merging

### Testing Fix 1: Email Shortening Logic

Run the new tests.

```bash
vendor/bin/phpunit tests/Logic/TestUsersHelper.php
```

**Manual Testing Steps**:
1. Create a user with a the same email address but different unique identifiers:

```php
   foreach ( range( 1, 10 ) as $i ) {
	$created_users[] = UsersHelper::create_or_get_user(
		[
			'user_login' => 'test' . $i,
			'user_email' => 'test@example.com',
		],
		'test' . $i 
	);
}
```

3. Check the email of the users created that have appended numbers like this

```
1test@example.com
21test@example.com
321test@example.com
...
```

4. Pull this PR and rerun the same code, and notice the appended numbers are fixed:

```
1test@example.com
2test@example.com
3test@example.com
...
```

### Testing Fix 2: Unique Identifier Only Lookup

Run the specific test for user lookup behavior:

**Manual Testing Steps**:
1. Create a user with specific data:
   ```php
   $user1 = UsersHelper::create_or_get_user([
       'user_email' => 'test@example.com',
       'user_login' => 'testuser',
       'display_name' => 'Test User'
   ], 'unique-123');
   ```

2. Try to create another user with the same email but a different unique identifier:
   ```php
   $user2 = UsersHelper::create_or_get_user([
       'user_email' => 'test@example.com',
       'user_login' => 'testuser',
       'display_name' => 'Test User'
   ], 'unique-456');
   ```

5. Verify that a new user is created (with a different email address)

**Expected Behavior**:
- Users with different unique identifiers should always create separate accounts
- Same unique identifier should always return the same user (regardless of other data)
- No fallback to email/login/nicename matching should occur

### DIFF
--- a/src/Logic/UsersHelper.php
+++ b/src/Logic/UsersHelper.php
@@ -232,10 +232,6 @@ class UsersHelper {
 
 		// First try with the uniqid for the user.
 		$wp_user = self::get_user_by_unique_identifier( $unique_identifier );
-		if ( ! $wp_user ) {
-			// OK, no unique identifier found, let's try to find the user by the data.
-			$wp_user = self::get_user( $data );
-		}
 		if ( $wp_user ) { // Great – we already have the user!
 			if ( ! empty( $data['role'] ) ) {
 				// If the role was passed in the data array – add it before returning.
--- a/src/Logic/UsersHelper.php
+++ b/src/Logic/UsersHelper.php
@@ -150,20 +150,23 @@ class UsersHelper {
 	public static function get_unused_fake_email( string $desired_email ): string {
 		$original_email = $desired_email;
 		if ( strlen( $desired_email ) > 100 ) {
-			// If the email is too long, we'll peel off a couple of characters from the beginning.
-			$desired_email = substr( $desired_email, 4 );
+			// If the email is too long, we'll peel off characters till we get 96 characters.
+			$email_parts   = explode( '@', $desired_email );
+			$desired_email = substr( $email_parts[0], 0, ( 96 - strlen( $email_parts[1] ) - 1 ) ) . '@' . $email_parts[1];
 			FileLog::get_logger( 'UsersHelper' )->warning( sprintf( 'Shortened email to under 100 chars from "%s" to "%s".', $original_email, $desired_email ) );
 		}
 
+		$generated_email = $desired_email;
+
 		$i = 0;
-		while ( false !== get_user_by( 'email', $desired_email ) ) {
-			$desired_email = ( ++$i ) . $desired_email; // Prepend.
+		while ( false !== get_user_by( 'email', $generated_email ) ) {
+			$generated_email = ( ++$i ) . $desired_email; // Prepend.
 		}
 		if ( $i > 0 ) {
-			CliLog::get_logger( 'UsersHelper' )->debug( sprintf( 'Generated fake email: %s.', $desired_email ) );
+			CliLog::get_logger( 'UsersHelper' )->debug( sprintf( 'Generated fake email: %s.', $generated_email ) );
 		}
 
-		return $desired_email;
+		return $generated_email;
 	}
 
 	/**
--- a/tests/Logic/TestUsersHelper.php
+++ b/tests/Logic/TestUsersHelper.php
@@ -157,4 +157,97 @@ class TestUsersHelper extends WP_UnitTestCase {
 		$should_also_be_shorter_and_different = UsersHelper::get_unused_nicename( $long_nicename );
 		$this->assertTrue( strlen( $should_also_be_shorter_and_different ) <= $max_length );
 	}
+
+	/**
+	 * Comprehensive test for get_unused_fake_email method.
+	 *
+	 * Tests various scenarios including the bug fix where long emails were not properly handled
+	 * when generating unique emails with prepended numbers.
+	 */
+	public function test_get_unused_fake_email_comprehensive() {
+		// Test 1: Basic functionality - email that doesn't exist
+		$test_email = 'test@example.com';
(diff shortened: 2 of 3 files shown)


src/Logic/UsersHelper.php
```diff
@@ -150,20 +150,23 @@ public static function get_unused_nicename( string $desired_nicename ): string {
...
 		if ( strlen( $desired_email ) > 100 ) {
-			// If the email is too long, we'll peel off a couple of characters from the beginning.
-			$desired_email = substr( $desired_email, 4 );
+			// If the email is too long, we'll peel off characters till we get 96 characters.
+			$email_parts   = explode( '@', $desired_email );
+			$desired_email = substr( $email_parts[0], 0, ( 96 - strlen( $email_parts[1] ) - 1 ) ) . '@' . $email_parts[1];
 			FileLog::get_logger( 'UsersHelper' )->warning( sprintf( 'Shortened email to under 100 chars from "%s" to "%s".', $original_email, $desired_email ) );
 		}
 
+		$generated_email = $desired_email;
```
naxoc, line 159: I think I was the one that introduced the `$original_email` var, but now that I look at it it gets confusing. What do you think of something like this where we keep the $desired_email as the original and then generate from that:
```php
		$generated_email = $desired_email;
		if ( strlen( $desired_email ) > 100 ) {
			// If the email is too long, we'll peel off characters till we get 96 characters.
			$email_parts   = explode( '@', $desired_email );
			$desired_email = substr( $email_parts[0], 0, ( 96 - strlen( $email_parts[1] ) - 1 ) ) . '@' . $email_parts[1];
			FileLog::get_logger( 'UsersHelper' )->warning( sprintf( 'Shortened email to under 100 chars from "%s" to "%s".', $desired_email, $generated_email ) );
		}

		$i = 0;
		while ( false !== get_user_by( 'email', $generated_email ) ) {
			$generated_email = ( ++$i ) . $desired_email; // Prepend.
		}
		if ( $i > 0 ) {
			CliLog::get_logger( 'UsersHelper' )->debug( sprintf( 'Generated fake email: %s.', $generated_email ) );
		}

		return $generated_email;
	}
```

tests/Logic/TestUsersHelper.php
```diff
@@ -157,4 +157,61 @@ public function test_too_long_unused_nicename() {
 		$should_also_be_shorter_and_different = UsersHelper::get_unused_nicename( $long_nicename );
 		$this->assertTrue( strlen( $should_also_be_shorter_and_different ) <= $max_length );
 	}
+
+	/**
+	 * Test that create_or_get_user no longer falls back to looking for users by other fields
```
naxoc: Does it make sense to have "no longer" in a test? After some time has passed any reference to when "no longer" was is gone. What do you think? Could we have the comment just explain what we are testing in the test?

naxoc (approved):
Looks good. I added some comments, but feel to free to merge if you disagree.

src/Logic/UsersHelper.php
kariae: This won't work because if we insert a long email (> 100 characters), it will return an email address that is still longer than 100 characters. You can run the `test_unused_email` to confirm this. We need three variables to store the original email (for the logging line), the truncated email (to ensure we're under 100 characters), and the attempted email addresses. I'm not a fan of the implementation as it's a bit complicated, but for now it works :D

tests/Logic/TestUsersHelper.php
kariae: Makes sense. I pushed a fix. What do you think?

src/Logic/UsersHelper.php
naxoc: All good. We can keep it as you have it :)

I think botched the example 🤣 The line that says `$desired_email = substr( $email_parts[0], 0, ( 96 - strlen( $email_parts[1] ) - 1 ) ) . '@' . $email_parts[1];` should be `$generated_email = substr( $email_parts[0], 0, ( 96 - strlen( $email_parts[1] ) - 1 ) ) . '@' . $email_parts[1];`

naxoc (approved):
I added a comment about my botched example. Feel free to merge or test that out :)

PR #92, opened 2025-07-10
Add PostUpdateHook

jorostoyanov:
## Overview

Occasionally, we need to update a Post but without updating the `post_modified` and `post_modified_gmt` post fields.

By using `wp_update_post`, this cannot be prevented by default. The alternative was to use a direct `$wpdb->update` call, but by doing so, we are missing a lot of extra features around `wp_update_post`.

Those include:
* having an automatically generated Revision for the updated post
* making sure that the `post_name` is unique in the `post_type`

The `PostUpdateHook` class in this PR solves the initial issue by hooking a function just before the post is updated and setting the `post_modified` and `post_modified_gmt` fields to the current ones.

## API

The `\Newspack\MigrationTools\Hooks\PostUpdateHook` class contains three static methods.

`attach_hook_to_prevent_modified_date_update()`

This method registers the callback. Make sure to call it before using `wp_update_post`.

---

`detach_hook_to_prevent_modified_date_update()`

This method unregisters the callback. Make sure to call it after using `wp_update_post`.

---

`fallback_to_current_post_modified_datetime()`

This method is the callback used that handles using the original `post_modified` and `post_modified_gmt` fields.

## How to Use

You can use the following snippet and replace the placeholders with some real data.

```php
\Newspack\MigrationTools\Hooks\PostUpdateHook::attach_hook_to_prevent_modified_date_update();

wp_update_post(
	[
		'ID' => {POST_ID},
		'post_content' => {POST_CONTENT},
	]
);

\Newspack\MigrationTools\Hooks\PostUpdateHook::detach_hook_to_prevent_modified_date_update();
```

## Caveats

👉 Make sure the post that is being updated supports revisions.
👉 When the updated post doesn't have Revisions prior to the update, the first revision will be the latest (updated) Post Content, so you may not see the Revisions Count in WP Admin.

## Automated Tests

// TODO

### DIFF
--- a/src/Hooks/PostUpdateHook.php
+++ b/src/Hooks/PostUpdateHook.php
@@ -3,48 +3,48 @@
 namespace Newspack\MigrationTools\Hooks;
 
 class PostUpdateHook {
-    /**
-     * Attaches a callback to prevent updating post_modified and post_modified_gmt
-     * when a post is being updated.
-     * 
-     * @return void
-     */
-    public static function attach_hook_to_prevent_modified_date_update(): void {
-        add_filter( 'wp_insert_post_data', [ __CLASS__, 'fallback_to_current_post_modified_datetime' ], 10, 4 );
-    }
+	/**
+	 * Attaches a callback to prevent updating post_modified and post_modified_gmt
+	 * when a post is being updated.
+	 * 
+	 * @return void
+	 */
+	public static function attach_hook_to_prevent_modified_date_update(): void {
+		add_filter( 'wp_insert_post_data', [ __CLASS__, 'fallback_to_current_post_modified_datetime' ], 10, 4 );
+	}
 
-    /**
-     * Detaches a callback to prevent updating post_modified and post_modified_gmt
-     * when a post is being updated.
-     * 
-     * @return void
-     */
-    public static function detach_hook_to_prevent_modified_date_update(): void {
-        remove_filter( 'wp_insert_post_data', [ __CLASS__, 'fallback_to_current_post_modified_datetime' ], 10 );
-    }
+	/**
+	 * Detaches a callback to prevent updating post_modified and post_modified_gmt
+	 * when a post is being updated.
+	 * 
+	 * @return void
+	 */
+	public static function detach_hook_to_prevent_modified_date_update(): void {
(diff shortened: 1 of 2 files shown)


iuravic (approved):
I love this! It's so practical, simple, and it's a new layer of logic for our plugin, i.e. `./src/Hooks`. Thanks for doing this 🙇 

What I tried testing it for is directly in Gutenberg editor, saving/updating something where I didn't want it to be seen that I updated it. And it works nicely.

ronchambers:
@jorostoyanov - It would be cool to add a static function to [the posts helper](https://github.com/Automattic/newspack-migration-tools/blob/trunk/src/Logic/Posts.php) something like:

```
public static function newspack_wp_update_post( $data ) {
  \Newspack\MigrationTools\Hooks\PostUpdateHook::attach_hook_to_prevent_modified_date_update();
   wp_update_post($data); 
   \Newspack\MigrationTools\Hooks\PostUpdateHook::detach_hook_to_prevent_modified_date_update();
}
```

If something like that is done, then the usage becomes just one line:

```
Newspack\MigrationTools\Logic\Posts::newspack_wp_update_post( $data );
```

I think something like this was discussed in Slack :-)

jorostoyanov:
Thanks @ronchambers !

Yes, this makes a lot of sense and I will add it in the next commit 💯

kariae:
@jorostoyanov, @ronchambers, I'm merging this one. It's much needed 🎉

jorostoyanov:
> @jorostoyanov, @ronchambers, I'm merging this one. It's much needed 🎉

Thanks, @kariae!

Let's discuss how revisions work when you use this approach to update posts :)

ronchambers:
I learned that WordPress don't save the revision until _after_ the post is updated, but I need a before the update revision.  So I'm doing the following so a revision will be saved prior to the update:

```
// Save revision before update.
wp_save_post_revision( $post->ID );

// Do replacement in db without updating the modified date.
$postarr = [
	'ID'           => $post->ID,
	'post_content' => $post->post_content,
];

$updated = Posts::update_post_without_modified_date( $postarr, true );

if( is_wp_error( $updated ) || $updated !== $post->ID ) {
	$this->logger->error( 'update error: ' . json_encode( $updated ) );
	exit();
}
```

So a before update revision will be saved, along with WordPress doing it's own update revision too.

See function added to: https://github.com/Automattic/newspack-migration-tools/pull/131

PR #96, opened 2025-07-17
Attachments: Prevent image sizes creation

jorostoyanov:
## Overview

This PR adds a filter to prevent creating new image sizes when importing external images.

## How to Test

* Upload an image
* Check the uploads folder
* There shouldn't be any cropped images except for `image.{jpg}` and `image-scaled.{jpg}`

### DIFF
--- a/src/Logic/Attachments.php
+++ b/src/Logic/Attachments.php
@@ -54,6 +54,9 @@ class Attachments {
 	 * @return int|WP_Error Attachment ID.
 	 */
 	public static function import_external_file( $path, $title = null, $caption = null, $description = null, $alt = null, $post_id = 0, $args = [], $desired_filename = '', $try_existing = true ) {
+		// Prevent creation of additional images.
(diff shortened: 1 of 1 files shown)


iuravic:
@jorostoyanov I think we should have this functionality available in our repo, when/if needed. Perhaps since we're quite busy with migrations ATM, and if there's not enough headroom to discuss proper integration, perhaps the safest an most painless way would be to make it into another Hook like in https://github.com/Automattic/newspack-migration-tools/pull/92. That way we can merge it safely right away so that this doesn't get lost, and once we decide further we can integrate it more.

naxoc:
I agree with Ivan that this filter should be in Atomic-specific code.

kariae:
@jorostoyanov, I agree with Camilla and Ivan here. The core WP expect to exists. I suggest if we have a viable case for this to add an optional parameter to the command instead and activate the filter if we pass that param, otherwise we can close this one.

naxoc:
@jorostoyanov What do you say? Can we close this one or should we try and find a place for it?

PR #99, opened 2025-07-21
Fix for get_or_create_category

iuravic:
Do not skip checking if taxonomy with same name and parent exists if unique identifier is used.

The `else` part of the ternary operator gets completely skipped if `$unique_identifier` is used, which would return an error if same category with name and parent exists.

### DIFF
--- a/src/Logic/Taxonomy.php
+++ b/src/Logic/Taxonomy.php
@@ -246,11 +246,17 @@ class Taxonomy {
 		$cat_description = $data['category_description'] ?? '';
 		$cat_nicename    = $data['category_nicename'] ?? '';
 
-		// Get term_id if it exists.
-		$existing_term_id = $unique_identifier
-			? $this->get_term_id_by_unique_identifier( self::UNIQUE_CATEGORY_IDENTIFIER_META_KEY, $unique_identifier )
-			: $this->get_term_id_by_taxonmy_name_and_parent( 'category', $cat_name, $cat_parent_id );
+		// Get term_id if it exists by $unique_identifier.
+		$existing_term_id = null;
+		if ( ! is_null( $unique_identifier ) ) {
+			$existing_term_id = $this->get_term_id_by_unique_identifier( self::UNIQUE_CATEGORY_IDENTIFIER_META_KEY, $unique_identifier );
+		}
+		// Get term_id if it exists by other data: $cat_name and $cat_parent_id.
+		if ( is_null( $existing_term_id ) ) {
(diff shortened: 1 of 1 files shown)


kariae:
@iuravic 👋 is this PR trying to fix creation of a category that has the same name of one that exists but with a different unique_idenfier?

iuravic:
> @iuravic 👋 is this PR trying to fix creation of a category that has the same name of one that exists but with a different unique_idenfier?

@kariae , kind of yes, but perhaps just a little bit different scenario:
- `$unique_identifier` is provided to `get_or_create_category`, together with `$data`
- category is not found by `$unique_identifier`, but that same category exists in DB by `$data` -- in my example, `$data` contains a category name and a category parent
- the ternary failed to get the existing category by `$unique_identifier`, and completely skipped trying to get it by name and parent
- the method returned `null`, signifying that category didn't exist, while it did

This splits ternary into two ifs, just so that if not found by `$unique_identifier`, it can still be searched by name and parent, and returned if exists.

Now that I think about it, perhaps this is supposed to happen, if category by `$unique_identifier` does not exist -- if this method is supposed to work exclusively with `$unique_identifier` -- and that's the primary thing it cares about, and `$data` is secondary? In that case do we need better handling in Foundation migration, or is everything OK and supposed to return an error?

kariae:
> the ternary failed to get the existing category by $unique_identifier, and completely skipped trying to get it by name and parent

That's a feature and not a bug 🤣 When using the `unique-_identifier`, we should rely solely on it to fetch the existing category. If we want to retrieve the category by name and parent, we should not pass a `unique_identifier` (though the name of the parameter).

We should not break this behavior, and the migration you are running is the perfect example why.

In the migration you are currently running, they had a few taxonomies. Section, Category, and Topics, with Section being migrated to a Parent Category in WordPress, Category migrated to a WordPress category with a parent, and Topic migrated to a WordPress category without a parent. Let's take the following Post as an example:

```JSON
// Sections.
{
    "oid": 1,
    "title": "Music"
},
{
    "oid": 2,
    "title": "Events"
},
// Categories
{
    "oid": 1,
    "title": "Rock",
    "section": 1
},
// Topics
{
    "oid": 100,
    "title": "Music",
},
```

The Topic `Music` can't be migrated to a WP category because it doesn't have a parent, and we already have a category named `Music`. This is something we should fix at the source.

iuravic:
@kariae , I understand what you mean. Makes sense now, it's all about the unique identifier here. Sorry for the confusion. Closing this PR.

PR #101, opened 2025-07-23
Add MemoryCleanupHook

jorostoyanov:
## Overview

This PR adds a `MemoryCleanupHook`.

This class allows running a migration on thousands of Posts, by cleaning the memory after on batches.

## How to Test

* Write a WP CLI migration
* Call `MemoryCleanupHook::cleanup` after processing e.g., 300 Posts.
* You should see a drop of memory usage compared to without cleaning the memory

## Example (Pseudo-code)

```
use Newspack\MigrationTools\Hooks\MemoryCleanupHook;

public function cmd_migration() {
    $page = 0;

    do {
        $posts = get_posts( [
            'posts_per_page' => 100,
            'paged' => $page,
        ] );

        foreach ( $posts as $post ) {
            // Processing Post.
        }

        MemoryCleanupHook::cleanup();

        $page++;
    } while ( count( $posts ) );
}
```

### DIFF
--- a/src/Hooks/MemoryCleanupHook.php
+++ b/src/Hooks/MemoryCleanupHook.php
@@ -9,13 +9,25 @@ class MemoryCleanupHook {
 	 * @static
 	 * @access public
 	 * 
-	 * @param int $sleep_time
+	 * @param int $sleep_time     Number of seconds to sleep between each flush.
+	 * @param int $current_step   Current counter/step. If provided $current_step and $flush_interval, will only flush every $flush_interval steps.
+	 * @param int $flush_interval Number of steps to wait before flushing again.
 	 */
-	public static function cleanup( int $sleep_time = 3 ): void {
-		self::reset_local_object_cache();
-		self::reset_db_query_log();
+	public static function cleanup( int $sleep_time = 0, int $current_step = null, int $flush_interval = null ): void {
 
-		sleep( $sleep_time );
+		// Determine if we should perform cleanup based on whether interval and step are provided.
+		$should_cleanup = ( is_null( $current_step ) || is_null( $flush_interval ) )
+			? true
+			: ( 0 === ( $current_step % $flush_interval ) );
+
+		if ( $should_cleanup ) {
+			self::reset_local_object_cache();
+			self::reset_db_query_log();
+
+			if ( $sleep_time > 0 ) {
+				sleep( $sleep_time );
+			}
+		}
 	}
 
 	/**
@@ -42,9 +54,15 @@ class MemoryCleanupHook {
 
 		foreach ( $properties as $property ) {
 			if ( property_exists( $wp_object_cache, $property ) ) {
-				$wp_object_cache->$property = [];
+				// Only set if the property is actually declared (not a magic property).
+				$reflection = new \ReflectionObject( $wp_object_cache );
+				if ( $reflection->hasProperty( $property ) ) {
+					$wp_object_cache->$property = [];
+				}
 			}
 		}
+
+		gc_collect_cycles();
 
 		if ( method_exists( $wp_object_cache, '__remoteset' ) ) {
 			$wp_object_cache->__remoteset(); // important
@@ -61,6 +79,10 @@ class MemoryCleanupHook {
 	 */
 	public static function reset_db_query_log(): void {
 		global $wpdb;
+
+		unset( $wpdb->queries );
+
+		gc_collect_cycles();
 
 		$wpdb->queries = [];
 	}
--- a/src/Hooks/MemoryCleanupHook.php
+++ b/src/Hooks/MemoryCleanupHook.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace Newspack\MigrationTools\Hooks;
+
+class MemoryCleanupHook {
+    /**
+     * Cleanup memory.
+     * 
+     * @static
+     * @access public
+     * 
+     * @param int $sleep_time
+     */
+    public static function cleanup( int $sleep_time = 3 ): void {
+        self::reset_local_object_cache();
+        self::reset_db_query_log();
+
+        sleep( $sleep_time );
+    }
+
+    /**
+     * Reset the local WordPress object cache
+     *
+     * This only cleans the local cache in WP_Object_Cache, without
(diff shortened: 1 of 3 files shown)


iuravic:
@jorostoyanov thanks for putting this together! 🙌  I've started running some tests now to measure the memory consumption without and with this cleanup, and will share when they're done running (ETA tomorrow/Monday).

jorostoyanov:
Thanks @iuravic 🙌 

I get the best results with batches of around 100 Posts.

jorostoyanov:
Hey @iuravic 👋 

Not trying to push here. Just excited to hear if you have any observations already 😅

src/Hooks/MemoryCleanupHook.php
```diff
@@ -0,0 +1,67 @@
...
+	 * @access public
+	 * 
+	 * @param int $sleep_time
+	 */
+	public static function cleanup( int $sleep_time = 3 ): void {
+		self::reset_local_object_cache();
+		self::reset_db_query_log();
+
+		sleep( $sleep_time );
+	}
```
iuravic: Proposing that we add the ability to only flush on certain interval. Added two more parameters,
- current step
- flush every custom number of steps -- interval

```
	 * @param int $sleep_time     Number of seconds to sleep between each flush.
	 * @param int $current_step   Current counter/step. If provided $current_step and $flush_interval, will only flush every $flush_interval steps.
	 * @param int $flush_interval Number of steps to wait before flushing again.
	 */
	public static function cleanup( int $sleep_time = 0, int $current_step = null, int $flush_interval = null ): void {
		
		// Determine if we should perform cleanup based on whether interval and step are provided.
		$should_cleanup = ( is_null( $current_step ) || is_null( $flush_interval ) ) 
			? true 
			: ( 0 === ( $current_step % $flush_interval ) );
		
		if ( $should_cleanup ) {
			self::reset_local_object_cache();
			self::reset_db_query_log();
			
			if ( $sleep_time > 0 ) {
				sleep( $sleep_time );
			}
		}
	}
```

src/Hooks/MemoryCleanupHook.php
```diff
@@ -0,0 +1,67 @@
...
+		$properties = [
+			'group_ops',
+			'memcache_debug',
+			'cache',
+		];
+
+		foreach ( $properties as $property ) {
+			if ( property_exists( $wp_object_cache, $property ) ) {
+				$wp_object_cache->$property = [];
+			}
```
iuravic: Though this looked good to me, when I ran it, I kept getting "Deprecated: Creation of dynamic property WP_Object_Cache::$memcache_debug is deprecated in /tmp/staging_full/MemoryCleanupHook.php on line 68".

It appears that happens because `property_exists()` is actually setting a property $memcache_debug on the $wp_object_cache object which doesn't exist. In other words, `property_exists()` doesn't work well with dynamic or magic properties, and still sets them.

Reflection on the other hand, properly deals with dynamic and magic properties, and this check is a bit sturdier and removes warnings.

```
				// Only set if the property is actually declared (not a magic property).
				$reflection = new \ReflectionObject( $wp_object_cache );
				if ( $reflection->hasProperty( $property ) ) {
					$wp_object_cache->$property = [];
				}
```

src/Hooks/MemoryCleanupHook.php
```diff
@@ -0,0 +1,67 @@
...
+		}
+
```
iuravic, line 64: I suggest that we should add one line here, after the foreach loop which sets empty arrays on properties:

```
gc_collect_cycles();
```

It will not do a drastic difference in PHP, but it may help a little.

Just running this one line like this, no arguments, right after the foreach, before the "// important" final if, that will be just fine. That's how I've been test-running it on NNE and Foundation sites.

src/Hooks/MemoryCleanupHook.php
```diff
@@ -0,0 +1,67 @@
...
+	 * Resets the WordPress DB query log.
+	 * 
+	 * @static
+	 * @access public
+	 * 
+	 * @return void
+	 */
+	public static function reset_db_query_log(): void {
+		global $wpdb;
+
```
iuravic, line 82: I propose we add these two lines between `global` declaration and the one line,
```
		unset( $wpdb->queries );
		gc_collect_cycles();
```

An extra unset might trigger a bit more active garbage collection, and another call to `gc_collect_cycles();` might help a little.

This is based on my tests on NNE and Foundation migrators the past couple of days, and works well.

iuravic (changes requested):
This is awesome!

I have tested and really used this hook during the past couple of days on NNE migrator, and on Foundation migrator.

I have proposed and a few tweaks,
- added the ability to auto-flush on custom interval/number of steps
- namely fixed one warning on dynamic/magic property access
- added a bit more explicit garbage collection -- does not really make a huge difference, but may help a little bit

During my testing -- where I covered all parts of this hook's code -- here's what I found:
1. On NNE migrator -- this migrator was not very demanding, everything worked well
  - sleeping or not did not matter
  - extra GC call did not matter, or mattered very little
  - on max 3000 cycles/steps/foreach-loops, every 100-500 steps worked best
3. On Foundation migrator -- this one showed a bit more preference, and worked well with more flushing
  - 100 cycles still resulted in memory running out, 50 cycles worked well
  - sleeping 3 every 50 cycles did the trick

**=> Review of findings:**
- migrators will be different and require different tuning
- the more demanding one worked well with flushing every 50 cycles and a little bit of sleep, too

jorostoyanov:
That sounds great, @iuravic 🙌 

Do you mind adding those changes to the branch, and we can probably merge this and have the rest of the Team to test it ☺️

iuravic:
> Do you mind adding those changes to the branch, and we can probably merge this and have the rest of the Team to test it 

Absolutely, @jorostoyanov , I'll push a commit now 🙇

iuravic:
Now pushed https://github.com/Automattic/newspack-migration-tools/pull/101/commits/c4f1ca8d0b68e7c734bf3151d1974f32a098712d with suggestions from review chat.

iuravic (approved):
Approved by L&I Team during Mad Dash.

PR #106, opened 2025-08-04
Proper delcaration of nullable arguments

iuravic:
Properly declaring nullable arguments to avoid deprecated warning in 8.1.


jorostoyanov (approved):
LGTM 👌

PR #109, opened 2025-08-05
Add JSONWriter Util class

jorostoyanov:
## Summary

Adds a `JSONWriter` utility class for writing JSON files.

## Purpose

The `JSONWriter` class makes it easy to write data to JSON files with minimal setup.

## Usage Example

```php

use Newspack\MigrationTools\Util\JSONWriter;

try {
    $json = new JSONWriter( 'output.json' ); // The file is created in the current working directory
    $json->put( [
        'name' => 'Alice',
        'email' => 'alice@example.com',
        'age' => 30,
    ] );
    $json->put( [
        'name' => 'Bob',
        'email' => 'bob@example.com',
        'age' => 23,
    ] );
    $json->close();
} catch ( Exception $e ) {
    error_log( $e->getMessage() );
}
```

## Tests

Unit tests are included in `tests/Util/JSONWriterTest.php` to cover file creation, row writing, and error conditions.

### DIFF
--- a/src/Util/JsonWriter.php
+++ b/src/Util/JsonWriter.php
@@ -52,16 +52,16 @@ class JsonWriter {
 	 */
 	public function put( array $item ): void {
 		if ( $this->is_first_item ) {
-            // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fwrite
+			// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fwrite
 			fwrite( $this->file_pointer, "[\n" );
 		} else {
-            // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fwrite
+			// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fwrite
 			fwrite( $this->file_pointer, ",\n" );
 		}
 
 		$json = wp_json_encode( $item, $this->flags );
 
-        // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fwrite
+		// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fwrite
 		fwrite( $this->file_pointer, "$json" );
 
 		$this->is_first_item = false;
@@ -74,7 +74,7 @@ class JsonWriter {
 	 * @throws Exception If the file cannot be closed.
 	 */
 	public function close(): void {
-        // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fwrite
+		// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fwrite
 		fwrite( $this->file_pointer, "\n]\n" );
 
 		if ( false === fclose( $this->file_pointer ) ) {
--- a/src/Util/JsonWriter.php
+++ b/src/Util/JsonWriter.php
@@ -72,6 +72,11 @@ class JsonWriter {
 	/**
 	 * Writes an object to the JSON file.
 	 * 
+	 * Handles three scenarios:
+	 * 1. First item in a new file: writes opening bracket and item
+	 * 2. First item when appending to existing file: removes closing bracket, adds comma, writes item
+	 * 3. Subsequent items: adds comma and writes item
+	 * 
 	 * @param  array $item The object data to write.
 	 * @return void
 	 * @throws Exception If the object cannot be written to the file.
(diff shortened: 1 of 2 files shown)


src/Util/JsonWriter.php
```diff
@@ -0,0 +1,94 @@
...
+		if ( false === fclose( $this->file_pointer ) ) {
+			// phpcs:ignore WordPress.Security.EscapeOutput.ExceptionNotEscaped
+			throw new Exception( "Could not close file: {$this->filename}" );
+		}
+	}
+
+	/**
+	 * Handles proper file ending when the object is destroyed.
+	 */
+	public function __destruct() {
```
naxoc, line 132: Love this! We should also add it to the CSVWriter :)

src/Util/JsonWriter.php
```diff
@@ -0,0 +1,94 @@
...
+	 * Closes the file pointer.
+	 * 
+	 * @return void
+	 * @throws Exception If the file cannot be closed.
+	 */
+	public function close(): void {
+		// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fwrite
+		fwrite( $this->file_pointer, "\n]\n" );
+
+		if ( false === fclose( $this->file_pointer ) ) {
```
naxoc, line 123: Should we just suppress this exception? I'm not sure what I would do with the handling if it failed. Log it?

naxoc (commented):
This looks good. It needs to have a git conflict resolved though. 

I added some small comments.

src/Util/JsonWriter.php
jorostoyanov: If the file cannot be closed, we might end up with corrupted data, and it's a good idea to know that before using it.

So yeah, a simple log is enough for me to know to be careful in the next step 😅

iuravic:
Howdy @jorostoyanov 👋  I checked out the class quickly, and here's a couple of smaller notes based on a first impression. I'm suggesting just a couple of comment additions and descriptions of methods, to clarify the usage, and maybe slight name changes.

---

When I first checked to figure out whether this class works with new or existing JSON files or both, it seems a bit ambiguous:
- the class opens files in 'a+' mode, which creates the JSON file if it doesn't exist, or appends to the end if it exists
- however, if it is given a new/empty file, it correctly produces a valid JSON array, which is OK, but if given an existing file it won't work correctly, it will append a new array to the end, creating invalid JSON like [...old content...][\n...new items...\n]\n. 
So I'm assuming it's made just to write new files only.

Could we please clarify this, add a note to the constructor maybe, so that it's clear it's only meant to work with new JSON files, not append to existing ones?

Instead of using a+, how about one of these two:
- 'w' writing only: truncates to zero length if the file exists; creates it if it doesn't exist
- 'x' exclusive write: creates the file and fails if it already exists (prevents accidental overwrites)

I would also propose perhaps to rename "JsonWriter" to reflect that it is only ment to write new files, not existing ones; How about "Json**SequentialFile**Writer" ? :)

---

`put` comment says "Writes an object to the JSON file." but it doesn't take an object, takes an array. Could we update that? I would suggest that we also explain how `put` actually works in the docblock, so that the user is aware what this is, maybe something like :
```
This method will append an array item directly to your JSON file:
- first call: writes [\n, then the JSON-encoded item
- subsequent calls: writes ,\n then the JSON-encoded item

Once putting is finished, it is necessary to run `close`.
```

Not sure if "first call" and "subsequent calls" bullets are necessary, perhaps not.

---

I would also recommend adding a description of `close`, maybe something like:
```
- writes \n]\n to close the array and completes the JSON formatting
- closes the file pointer

Even though if you don't call close() explicitly `__destruct()` will call
it when the object is destroyed, your file won't be correctly formatted without `close()`.
```

---

If I'm not mistaken, in case the user doesn't call `close`, the destructor will be called and close the file pointer. However the `fwrite( $this->file_pointer, "\n]\n" );` will not be written in that case and the resulting file's syntax will be unfinished and JSON contents invalid. Why not also fwrite the closing JSON syntax characters in the destructor, and in that case also end up with a valid JSON-syntax file?

iuravic (changes requested):
Howdy @jorostoyanov , I left a couple of small notes [here](https://github.com/Automattic/newspack-migration-tools/pull/109#issuecomment-3638528358) :)

jorostoyanov:
Hey @iuravic 👋 

Thank you for the detailed feedback!

I ran some tests, and with the help of Claude we updated the JsonWriter class to support existing files. Given that, I updated the docs to better reflect the existing features and added a new test to cover the existing file operation. Let me know hot it looks to you.

> If I'm not mistaken, in case the user doesn't call close, the destructor will be called and close the file pointer. However the fwrite( $this->file_pointer, "\n]\n" ); will not be written in that case and the resulting file's syntax will be unfinished and JSON contents invalid. Why not also fwrite the closing JSON syntax characters in the destructor, and in that case also end up with a valid JSON-syntax file?

The closing bracket is handled in the `close` method and it's being used in the `__destruct` method. Can you follow-up with an example on how it won't add the ending bracket?

Thanks! 🙌

jorostoyanov:
Hey @iuravic 👋 

Can you check the last comment when you have the time?

Thanks!

iuravic (approved):
Thank you @jorostoyanov ! I apologize for my super late reply, looks like my notifications got marked as done by mistake :(

This looks great, thanks for tweaking it a bit more.

> closing bracket
I agree, this looks fine to me 👍

PR #110, opened 2025-09-09
Fix Redirection Plugin Needs Setup for DB

ronchambers:
It's not enough to just check if the Redirection plugin is installed/active to use this helper.  The Redirection plugin's own setup method needs to be run before using this helper, otherwise the functions will not work properly, and errors will be written to the debug log.

This code makes sure the Redirection plugin's setup was run by checking the existence of one of the Redirection db tables.

As noted in the php comments:

```
// Setup can be performed in wp-admin > tools > Redirection ...
// or with cli command: wp redirection database install
```

**Developer notes:**

Install with CLI:

```
wp redirection database install
````

Or run the setup in WP-Admin:

redirection-setup

### DIFF
--- a/src/Logic/Redirection.php
+++ b/src/Logic/Redirection.php
@@ -16,8 +16,8 @@ class Redirection {
 			NMT::exit_with_message( 'The Redirection plugin ( redirection ) is a dependency, and will have to be installed and activated before this helper class can be used.' );
 		}
 
-		// Make sure the Redirections plugin is "setup" (ie: it's database tables were installed).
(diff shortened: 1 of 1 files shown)


ronchambers:
@kariae - please review when you can.  This is a simple change to make sure the Redirection plugin's database table exists so the helper will actually work properly and not just throw errors in the debug log.  Thanks!

kariae (approved):

kariae:
Ah, I always struggle with the Redirection plugin for the first few minutes before remembering that I need to activate it. Thank you for this PR!

PR #111, opened 2025-09-10
fix phpcs phpunit errors

ronchambers:
Github actions for both PHPCS and PHPUNIT were failing, so I fixed the problems.

1) PHPCS was failing due to a space in a `docblock` in `src/Hooks/MemoryCleanupHook.php`.

2) PHPCS was failing due to a missing true value in `get_post_meta` in `src/Command/PostFeaturedVideoMigrator.php`.

3) PHPUNIT was failing due to some version issues, so I did the following locally using:

```
PHP version 8.1.31
Composer version 2.2.6 
```
Steps:
```
rm composer.lock 
rm -rf vendor/
composer clearcache
composer install
```
An updated `composer.lock` was created and I committed it to this PR.

**Conclusion**: PHPCS and PHPUNIT both are working in Github actions below.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "php": ">=8.1",
         "monolog/monolog": "^3.7",
         "bramus/monolog-colored-line-formatter": "^3.1",
-        "halaxa/json-machine": "^1.1.5"
+        "halaxa/json-machine": "^1.1"
     },
     "require-dev": {
         "ext-libxml": "*",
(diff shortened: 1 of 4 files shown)


ronchambers:
Hi @iuravic - can you give this a review?  All three issues seem to be related to stuff you had worked on.  Can you also make sure this PR still allows your previous commit `Update halaxa/json-machine to 1.1.5 in order to support PHP 8.4` still works?  I'm not sure how to test that...  Thanks!

iuravic (approved):

PR #112, opened 2025-09-18
Fix Redirection query matching: Change to pass.

ronchambers:
Change the default value from "exact" to "pass" when using this helper.  

See write up in NP Engineering P2: search for: `/?s=redirection-plugin-query-string-parameters-matching-options`


ronchambers:
@kariae - when you have time to review, this PR addes a docblock to the Redirections helper with recommendation on why "pass" is the preferred option and also updates the default argument to "pass" instead of "exact" when creating Redirection urls.  Thanks!

kariae (approved):
Nice catch!

PR #114, opened 2025-09-26
NMT loggers 'debug' default level

iuravic:
Instead of "info", we should have "debug" as default level. When "info" was the default level, everything on CLI was always green, which kind of beats the purpose of any green highlighting. Additionally the gray "debug" output did not work at all. Now that "debug" is set as default, info and everything else still works as usual, we just have debug, too.

Memo -- available log levels, from lowest to highest:
- DEBUG
- INFO
- NOTICE
- WARNING
- ERROR
- CRITICAL
- ALERT
- EMERGENCY

### DIFF
--- a/src/NMT.php
+++ b/src/NMT.php
@@ -28,13 +28,13 @@ class NMT {
 	 * Private on purpose
 	 */
 	private function __construct() {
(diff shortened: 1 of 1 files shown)


iuravic:
Approved by the Team during the huddle, merging.

PR #116, opened 2025-10-03
Users Helper Assign Authors

ronchambers:
### Assigning Authors to a Post

The `assign_authors_to_post()` function can be used to assign author(s) to a post. 

```php

// Assign one author (single value array) to the post.
UsersHelper::assign_authors_to_post( 1, [ 1 ] );

// Assign multiple authors (multiple value array) to the post.
UsersHelper::assign_authors_to_post( 1, [ 1, 2, 3 ] );

```

View changed files to see: updated docs, new function (with docblock), and new test.

### DIFF
--- a/docs/users-helper.md
+++ b/docs/users-helper.md
@@ -40,6 +40,26 @@ Obviously – the more data you pass the better, but the method will create a us
 
 Fake emails will be created that look something like f3e895c937@example.com. 
 
+## Assigning Authors to a Post
+
+The `assign_authors_to_post()` function can be used to assign author(s) to a post. The [Co-Authors Plus](https://wordpress.org/plugins/co-authors-plus/) plugin is required to use this function.  (Co-Authors Plus is Newspack's preferred plugin for managing "multiple authors per post").
+
+Co-Authors Plus will create author taxonomy relationships in the `terms` db tables, and also update the post's `post_author` column too.
+
+### Usage
+
+```php
+
+// Assign one author (single value array) to the post.
+UsersHelper::assign_authors_to_post( 1, [ 1 ] );
+
+// Assign multiple authors (multiple value array) to the post.
+UsersHelper::assign_authors_to_post( 1, [ 1, 2, 3 ] );
+
+```
+
+Please see the underlying Co-Authors Plus function [add_coauthors](https://github.com/Automattic/Co-Authors-Plus/blob/30602dbd59c6cd73bd4aa3ff8a3e6eda0c1bccea/php/class-coauthors-plus.php#L1022) (specifically the variables [$query_type and $field](https://github.com/Automattic/Co-Authors-Plus/blob/30602dbd59c6cd73bd4aa3ff8a3e6eda0c1bccea/php/class-coauthors-plus.php#L1050-L1052)) for other argument options.
+
 ## Logging
 This class logs to file in `UsersHelper.log`.
 
--- a/src/Logic/GuestContributorsHelper.php
+++ b/src/Logic/GuestContributorsHelper.php
@@ -8,7 +8,7 @@ use WP_User;
 use WP_Error;
 use WP_Role;
 
-class GuestContributorsHelper extends UsersHelper {
+class GuestContributorsHelper {
 
 	const ERROR_ATTEMPTS        = 'Might be in an infinite loop.';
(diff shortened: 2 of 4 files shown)


ronchambers:
@kariae - please review when you can.  This PR adds the ability to assign authors to post directly using UsersHelper.  (I'm moving the GuestContributors stuff one PR at a time).

iuravic (changes requested):
Thank you @ronchambers , I tested it and it works beautifully. However this now looks like a duplicate of [GuestContributorsHelper::assign_contributors_to_posts](https://github.com/Automattic/newspack-migration-tools/blame/trunk/src/Logic/GuestContributorsHelper.php#L134).

Calling @kariae for opinion -- should we also remove the `GuestContributorsHelper::assign_contributors_to_posts` so that we only use this one? 

Also I'd propose we now deprecate or rather completely delete the old [`Newspack\MigrationTools\Logic\CoAuthorsPlusHelper`](https://github.com/Automattic/newspack-migration-tools/blob/trunk/src/Logic/CoAuthorsPlusHelper.php) class. That was the first version of co-author functionality ever, now old and still contains Guest Authors functionality. I don't think it's needed any longer with our new classes.

@ronchambers checking in about a couple of things:
- the `$query_type = 'id'` argument -- are we ever going to use this with a value different than `id`? If not, perhaps we can remove it from our wrapper method?
- let's please add type definition to `$append = false` to `bool $append = false` in function declaration?
- a tiny and optional suggestion -- I'd suggest we didn't use `const` for error messages. IMHO error messages are a bit more readable/searchable when used inline (typically they could even be translated on the fly that way, which we're not going to do). This is a personal preference, if you don't like it free to leave it

ronchambers:
> ...this now looks like a duplicate of GuestContributorsHelper::assign_contributors_to_posts ...
> Calling at-kariae for opinion -- should we also remove the `GuestContributorsHelper::assign_contributors_to_posts` so that we only use this one?

Correct, that is the old function that will become deprecated with [a PR here](https://github.com/Automattic/newspack-migration-tools/pull/117/files) after we merge this one.

--

> Also I'd propose we now deprecate or rather completely delete the old [`Newspack\MigrationTools\Logic\CoAuthorsPlusHelper`](https://github.com/Automattic/newspack-migration-tools/blob/trunk/src/Logic/CoAuthorsPlusHelper.php) class. That was the first version of co-author functionality ever, now old and still contains Guest Authors functionality. I don't think it's needed any longer with our new classes.

Yes, the old CoAuthorsPlusHelper will be removed/deprecated/etc after we get the UsersHelper and GuestContributors all finalized.  I'm doing small PRs one-by-one for easy to understand reviews.

--

> the `$query_type = 'id'` argument -- are we ever going to use this with a value different than `id`? If not, perhaps we can remove it from our wrapper method?

User ids would probably cover most cases, but since the underlying function `$coauthors_plus->add_coauthors` allows other types like user_nicename, user_login, user_email, etc, I think we should just leave the wrapper argument available just in case.

--

>  let's please add type definition to `$append = false` to `bool $append = false` in function declaration?

Will do.

--

> a tiny and optional suggestion -- I'd suggest we didn't use `const` for error messages. IMHO error messages are a bit more readable/searchable when used inline (typically they could even be translated on the fly that way, which we're not going to do). This is a personal preference, if you don't like it free to leave it

Cool, makes sense.  I'll review this idea.

kariae:
> Correct, that is the old function that will become deprecated with [a PR here](https://github.com/Automattic/newspack-migration-tools/pull/117/files) after we merge this one.

Can we make `assign_contributors_to_posts` call `UsersHelper::assign_authors_to_post` instead of deprecating for convenience? If it's not used anywhere, we can remove it.

> a tiny and optional suggestion -- I'd suggest we didn't use const for error messages. IMHO error messages are a bit more readable/searchable when used inline

I agree with the searchability point. @ronchambers let's do these small tweaks and merge this one! Awesome job. Thank you!

ronchambers:
All fixed.

- type definitions added
- error messages changed to searchable strings
- deprecation note moved to future deprecation pr ( https://github.com/Automattic/newspack-migration-tools/pull/117 )

kariae (approved):
LGTM. Thank you, Ron!

PR #117, opened 2025-10-07
Guest Contributors Deprecate Assign

ronchambers:
Now that we have `UsersHelper::assign_authors_to_post` [merged](https://github.com/Automattic/newspack-migration-tools/pull/116), we no longer need the assign function on Guest Contributors.

Zak had also asked: 

> Can we make `assign_contributors_to_posts` call `UsersHelper::assign_authors_to_post` instead of deprecating for convenience?  If it's not used anywhere, we can remove it. -- Zak (via https://github.com/Automattic/newspack-migration-tools/pull/116#issuecomment-3454957917 )

Yes.  This PR will show a deprecation warning but will also run the assign to post also so that nothing breaks.

Otherwise what we could do is just update NCCM at these spots ( [one](https://github.com/Automattic/newspack-custom-content-migrator/blob/5a62fa148eb40d346e69737964e2e55860d5fece/src/Command/General/Foundation/FoundationMigrator.php#L1313) and [two](https://github.com/Automattic/newspack-custom-content-migrator/blob/5a62fa148eb40d346e69737964e2e55860d5fece/src/Command/General/Foundation/FoundationMigrator.php#L1679) ) to change those to use the UsersHelper instead.

My preference is to just merge this PR as is.  There will be more Guest Contributors changes in the near future, so this is a small non-breaking change.

### DIFF
--- a/docs/guest-contributors.md
+++ b/docs/guest-contributors.md
@@ -104,28 +104,7 @@ if ( is_wp_error( $users ) ) WP_CLI::error( $users->get_error_message() );
 
 #### assign_contributors_to_post
 
-Assigns Guest Contributors to a Post using Co-Authors Plus.
-
-**Parameters:**
-
--   `$post_id` (int): Post ID.
--   `$contributor_ids` (array): Array of contributor IDs.
-
-**Returns:** `true|WP_Error` - True if successful, WP_Error if not.
-
-Example usage:
-
-```php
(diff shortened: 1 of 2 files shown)


ronchambers:
@kariae @iuravic - for review.  Thanks!

kariae (approved):
All makes sense to me!

Issue #77, opened 2025-06-11
Guest Contributors - Enhancements


ronchambers:
After the main PR ( https://github.com/Automattic/newspack-migration-tools/pull/49 ) is merged, do this:

- Check for email addresses in display name so that user_login, user_nicename, nickname, etc, don't end up being "myname_gmail_com" or something that could be used to hack the account.

- Keep the GuestContributorsHelper class, but move much of the underlying functionality into the UsersHelper class.  The GuestContributors would just call the UsersHelper with the guest contributors role. 

- Create an "assign to posts" function.  ([This is currently](https://github.com/Automattic/newspack-migration-tools/blob/deeefd47fd486820b5e6c9a72234cfb0d4cf22e5/docs/guest-contributors.md#to-assign-guest-contributors-to-posts-use-coauthors-plus) in the help doc, but make it an actual function). 

- Make a single `get_create_contributor` function.

- Related: https://github.com/Automattic/newspack-migration-tools/issues/41

- Trim `display_name` to 250 chars during data validation per this comment thread: https://github.com/Automattic/newspack-migration-tools/pull/119#issuecomment-3472266489 [PR](https://github.com/Automattic/newspack-migration-tools/pull/125)

PR #119, opened 2025-10-09
Users Helper: capture core bug on insert

ronchambers:
Additional error checking for when `wp_insert_user` returns integer 0 - which then causes `get_user_by( 'ID', $user_id )` to not be able to get a user.

This problem can happen in WP core on [these lines](https://github.com/WordPress/WordPress/blob/884b83238ef970816587d8f877adc9bf0fc5f55d/wp-includes/user.php#L2517-L2518).

`$wpdb->insert( $wpdb->users, $data );` may return `false` on error.  But instead of a `WP_Error` being returned, the next line is run `$user_id = (int) $wpdb->insert_id;` which causes `false` to become `0`.  Thus `wp_insert_user` will return `0` instead of a proper `WP_Error`.

One example of such case is this core bug [here](https://core.trac.wordpress.org/ticket/53109).

**It's important that whenever `wp_insert_user` is called, to not only check `is_wp_error( $user_id )` but also make sure the return value is an integer greater than 0.**

The odds of this bug happening are rare (like if `display_name` is over `250 characters`), but we should still capture it if it happens.

### DIFF
--- a/src/Logic/UsersHelper.php
+++ b/src/Logic/UsersHelper.php
@@ -415,10 +415,11 @@ class UsersHelper {
 			throw new Exception( sprintf( 'Could not create user: %s. Context data: %s', $user_id->get_error_message(), wp_json_encode( $data ) ) );
 		}
 		if ( ! ( $user_id > 0 ) ) {
-			// Core bug could result in "0" being returned: https://core.trac.wordpress.org/ticket/53109
-			// This happens if display_name is > 250 characters. A wp_error should be returned from wp_insert_user, but instead a 
-			// value of "0" is returned. We need to check for this case since get_user_by needs a $user_id > 0, otherwise $wp_user will equal "false".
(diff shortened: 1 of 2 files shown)


ronchambers:
FYI: @kariae @iuravic @jorostoyanov @naxoc @eddiesshop - always verify the return value when doing `wp_insert_user()` - see above.

naxoc:
I love this check! But I wonder if it would be better to make sure that the display name is never more than 250 chars? I'm not sure we always will remember the exception handling and shorten the name in the calling code?

kariae:
> it would be better to make sure that the display name is never more than 250 chars

I was thinking about the same thing. We're already doing some sanitization for user data, we can add this too.

ronchambers:
Merged.  I also captured the task to trim `display_name` on my enhancements issue: https://github.com/Automattic/newspack-migration-tools/issues/77

PR #121, opened 2025-10-15
Do not allow comma in Collection title

iuravic:
Does not allow comma in Collection titles.
More about the issue [here](https://a8c.slack.com/archives/C055SRT7WN4/p1752597885857479).

### DIFF
--- a/src/Logic/CollectionsHelper.php
+++ b/src/Logic/CollectionsHelper.php
@@ -82,7 +82,7 @@ class CollectionsHelper {
 		}
 
 		// Unfortunately, collection titles can not contain commas because of a known core issue, re: https://core.trac.wordpress.org/ticket/14691.
-		if ( false !== strpos( $data['post_title'], ',' ) ) {
+		if ( str_contains( $data['post_title'], ',' ) ) {
 			return new WP_Error( 'invalid_collection_title', 'Collection title cannot contain a comma.' );
 		}
 
--- a/src/Logic/CollectionsHelper.php
+++ b/src/Logic/CollectionsHelper.php
@@ -81,6 +81,11 @@ class CollectionsHelper {
 			return $wp_post_id;
 		}
 
+		// Unfortunately, collection titles can not contain commas because of a known core issue, re: https://core.trac.wordpress.org/ticket/14691.
+		if ( false !== strpos( $data['post_title'], ',' ) ) {
+			return new WP_Error( 'invalid_collection_title', 'Collection title cannot contain a comma.' );
+		}
+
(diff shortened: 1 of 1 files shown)


kariae:
@iuravic, the tagged issue is for tag names, while the Collections are a custom post type. Do you know if we noticed this issue somewhere?

src/Logic/CollectionsHelper.php
```diff
@@ -81,6 +81,11 @@ public function get_or_create_collection( array $data, string $unique_identifier
 			return $wp_post_id;
 		}
 
+		// Unfortunately, collection titles can not contain commas because of a known core issue, re: https://core.trac.wordpress.org/ticket/14691.
+		if ( false !== strpos( $data['post_title'], ',' ) ) {
```
jorostoyanov: Just a suggestion..

Can we use `str_contains` ([Docs](https://www.php.net/manual/en/function.str-contains.php)) to check for a comma? The function is new in PHP v8, and it's easier to read IMO.

```php
if ( str_contains( $data['post_title', ','] ) ) {
    ...
}
```

jorostoyanov (changes requested):
LGTM 👍 

Just a quick suggestion to use `str_contains` for readability. Ignore it if you don't agree :)

jorostoyanov:
> @iuravic, the tagged issue is for tag names, while the Collections are a custom post type. Do you know if we noticed this issue somewhere?

@kariae Collections use Collection Taxonomy as a hidden taxonomy behind the scenes, which allows relating Posts with Collections. Newspack plugin keeps Collections CPT with Collections Taxonomy in sync [here](https://github.com/Automattic/newspack-plugin/blob/trunk/includes/collections/class-sync.php).

src/Logic/CollectionsHelper.php
iuravic: @jorostoyanov , sure thing, it's definitely more readable, now updated here https://github.com/Automattic/newspack-migration-tools/pull/121/commits/0db7226a4d0e1388ddaa1e095a6c1a7f721946d2

jorostoyanov: Thank you! 🙇

jorostoyanov (approved):

kariae:
> @kariae Collections use Collection Taxonomy as a hidden taxonomy behind the scenes, which allows relating Posts with Collections. Newspack plugin keeps Collections CPT with Collections Taxonomy in sync [here](https://github.com/Automattic/newspack-plugin/blob/trunk/includes/collections/class-sync.php).

Yes, that makes total sense!

jorostoyanov (approved):

PR #125, opened 2025-11-08
Fix Users Helper Display Name sanitization

ronchambers:
### Fixes:

* Trim `display_name` to maximum database column length.
* Sanitize `display_name` so email addresses don't end up in publicly viewable Display Name.

### Notes:

* Sanitization happens early so that if `display_name` contained an email "@"; the email will not be passed along to other variables like `user_nicename`.
* A `display_name` like `Ron Chambers, News Reporter, ron@example.com` isn't exactly an email, but the "@" portion will still be trimmed.

### Errors / Tests:

Since we're now sanitizing `display_name` to the max database column length, the rare case of `wp_insert_user` returning integer "0" becomes even more rare.  ( [merged pr here](https://github.com/Automattic/newspack-migration-tools/pull/119) ).  But it's still important we check for `$user_id > 0` (since it could happen with other db columns).  Anyway...the comments around this code were reduced and the test was removed now that it's so rare a test isn't needed (and hard to produce).


ronchambers:
The PR above will now trim `display_name` to the db column max when creating Users, and also make sure Display Name (which is publicly visible) doesn't contain an email address.

@kariae @naxoc @iuravic 

Thanks!

kariae (approved):
LGTM! Thank you, Ron!

PR #126, opened 2025-11-10
Add an optional limit when getting terms assigned to max posts.

naxoc:
When querying for A LOT of tags with this we'd run out of memory. This makes the function able to use batching.

Test it by eyeballing it or calling it with a limit.


kariae (approved):
LGTM! Thank you!

PR #129, opened 2025-11-20
Reassign post authorship from user A to user B

iuravic:
New method `reassign_author` in `UsersHelper` which reassigns post authorship from one user to another. It makes sure that other coauthors remain properly assigned, and just changes that one author from A to B.

Note for our Team -- in [this P2](https://newspackengineering.wordpress.com/2025/09/10/problems-when-instantiating-helpers-in-constructors/#comment-2207) we agreed to check if dependencies plugins are active in the constructor, but that doesn't seem to work if we use static methods for everything. That's why this checks if CAP is active in the new method directly.

### DIFF
--- a/src/Logic/UsersHelper.php
+++ b/src/Logic/UsersHelper.php
@@ -546,15 +546,8 @@ class UsersHelper {
 	 *
 	 * @return bool|null|WP_Error True if reassignment was successful. Null if $from_user_id is not one of post's coauthors.
 	 *                            WP_Error no coauthors are found for post or assignment fails.
-	 *
-	 * @throws Exception If Co-Authors Plus plugin is not active.
 	 */
 	public static function reassign_author( int $post_id, int $from_user_id, int $to_user_id ): bool|null|WP_Error {
-		// Validate CAP is available.
-		if ( ! is_plugin_active( 'co-authors-plus/co-authors-plus.php' ) || ! function_exists( 'get_coauthors' ) ) {
-			throw new Exception( 'Co-Authors Plus plugin is not active.' );
-		}
-
 		// Get current authors for the post.
 		$current_authors = get_coauthors( $post_id );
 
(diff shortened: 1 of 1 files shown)


naxoc:
This looks good! I do wonder if it would make more sense in the `CoAuthorsPlusHelper` class though? Not sure where we are with that class now that we have guest authors, but I wanted to see what you think, Ivan?

iuravic:
For the record, review discussed in more detail in our [internal Slack thread](https://a8c.slack.com/archives/C013N9S2C7Q/p1763637469409689).

naxoc (approved):
Looks good! A small nit could be to use false instead of null for the return value to keep the checks simpler in the calling code, but that one is optional

iuravic:
> Looks good! A small nit could be to use false instead of null for the return value to keep the checks simpler in the calling code, but that one is optional

@naxoc thanks for this note, it's a really interesting point, I love it...! Whether to add more detailed info to return with the null, or simplify it to either true or WP_Error. To me it also seems that both would work well.

My thought behind putting the null here was that with it the post can be skipped silently, without logging any success or failure,
```
foreach ( $post_ids as $post_id ) {
  $result = UsersHelper::reassign_author( $post_id, $from_user_id, $to_user_id );
  if ( true === $result ) {
      $this->log( 'Updated.' );
  } elseif ( null === $result ) {
      // Silent skip - given user was not an author, which is neither good or bad, so can be just skipped silently
  } elseif ( is_wp_error( $result ) ) {
      $this->log( 'Houston, Houston, ... );
  }
}
```

So while it gives extra control, on the other hand it requires another check. I totally get that.

What I usually do in cases like this, even if there is an extra null, and I want to simplify the control in the wrapper is something like,
```
if ( true === $result ) {
  // Yey!
} else {
  $this->log( 'Something hit the fan. Error message: %s', is_wp_error( $result ) ? $result->get_error_message() : 'n/a' );
}
```

ronchambers:
@iuravic - I think i like the 3-way return...but then again I kinda lean toward just 2 returns of either true or WP_Error (with different error codes)...but the 3 way is kinda cool too.  Never thought about it.   I'll let it rattle around in my head for a future function :-)

* bool|null|WP_Error

PR #130, opened 2025-11-21
Fix/users helper cap validation

ronchambers:
This PR refactors the validation for Co-Authors Plus plugin (and it's functions) for use with some of our UsersHelper functions.

* Reinstates validation to `reassign_author` function since we need to check that CAP is active before calling it's function `get_coauthors` ( [re-instated](https://github.com/Automattic/newspack-migration-tools/pull/129/commits/502aa406636cfd485e9c534ed6c30cf470550310) concept from [PR](https://github.com/Automattic/newspack-migration-tools/pull/129) ) 
* Since our `CoAuthorsPlusHelper` will someday be fully deprecated/removed, all new functionality should be put into `UsersHelper`.  ( [old validation for reference](https://github.com/Automattic/newspack-migration-tools/blob/4df31617ea50caa59ffc09da2271f979cbd430f0/src/Logic/CoAuthorsPlusHelper.php#L47) ).
* It might be cool in the future to create some sort of trait or util that can be re-used to "validate helper dependencies".  For helpers that need to check if a plugin is installed or not, it would be cool to have a standard approach to this, but for now this PR puts the validation into a function in the same class.

### DIFF
--- a/src/Logic/UsersHelper.php
+++ b/src/Logic/UsersHelper.php
@@ -593,9 +593,9 @@ class UsersHelper {
 	
 		// Only run this function once, as long as argument(s) are the same.
 		static $validated = [];
-		$args_hash = wp_hash( serialize( func_get_args() ) );
-		if ( isset( $validated[$args_hash] ) ) {
-			return $validated[$args_hash];
+		$args_hash        = wp_hash( wp_json_encode( func_get_args() ) );
+		if ( isset( $validated[ $args_hash ] ) ) {
+			return $validated[ $args_hash ];
 		}
 
 		if ( ! function_exists( 'is_plugin_active' ) ) {
@@ -622,6 +622,8 @@ class UsersHelper {
 			return new WP_Error( 'ERROR_COAUTHORS_PLUS_GAS', 'Co-Authors Plus Guest Authors not set.' );
 		}
 
-		return ( $validated[$args_hash] = true );
+		$validated[ $args_hash ] = true;
+		
+		return $validated[ $args_hash ];
 	}
 }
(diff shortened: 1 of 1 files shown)


naxoc (commented):
This looks good! 

One small request: could we keep a static variable like `$validated` in the `validate_co_authors_plus` function and then only do the validation once and after that just return the var? No need to check all this for every author

ronchambers:
@naxoc - thanks, I added static `$validated` - for each argument set.

```
// Only run this function once, as long as argument(s) are the same.
static $validated = [];
$args_hash        = wp_hash( wp_json_encode( func_get_args() ) );
if ( isset( $validated[ $args_hash ] ) ) {
	return $validated[ $args_hash ];
}
```

src/Logic/UsersHelper.php
```diff
@@ -576,4 +579,51 @@ public static function reassign_author( int $post_id, int $from_user_id, int $to
...
+	 *
+	 * @param bool $guest_authors Is the Guest Authors feature of CAP required. Default is not required.
+	 * 
+	 * @return bool|WP_Error True if active, WP_Error if not.
+	 */
+	public static function validate_co_authors_plus( $guest_authors = false ): bool|WP_Error {
+	
+		// Only run this function once, as long as argument(s) are the same.
+		static $validated = [];
+		$args_hash        = wp_hash( wp_json_encode( func_get_args() ) );
```
naxoc: Sorry to keep going at this, but because this gets called a lot I think the hashing and encoding will also take up time too. We could simplify the array key like this maybe?
`$key = $guest_authors ? '1' : '0';` (`$key` can replace `$args_hash` here.)

ronchambers: Thanks!  I don't mind making it perfect ;-)  Done.

naxoc (approved):
Looks good! Thank you for the patience, Ron :)

PR #131, opened 2025-11-24
Update post without modified date, plus save revision

iuravic:
Method which updates a post without touching the modified date, but also saves a revision before.

It is @ronchambers idea 🙌 here putting it into a new method.

### DIFF
--- a/src/Logic/Posts.php
+++ b/src/Logic/Posts.php
@@ -792,9 +792,11 @@ SQL;
 
 	/**
 	 * Shorthand for updating a post without touching the post_modified and post_modified_gmt fields,
-	 * but it saves a revision before the update.
+	 * but instead of wp_update_post only doing a revison after the update, this functions saves a
+	 * revision before the update too.
 	 *
-	 * Parameters are the same as for wp_update_post.
+	 * Parameters are the same as for wp_update_post. Even though wp_update_post allows $postarr to be
+	 * optional, it's required here so that we have a post ID for saving the revision.
 	 *
 	 * @static
 	 * @access public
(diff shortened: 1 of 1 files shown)


src/Logic/Posts.php
```diff
@@ -789,4 +789,35 @@ public static function update_post_without_modified_date( $postarr = array(), $w
...
+	 * @see https://developer.wordpress.org/reference/functions/wp_update_post/
+	 *
+	 * @param array|object $postarr          Post data. Arrays are expected to be escaped, objects are not. Default array. ID key or property is required. @see wp_update_post().
+	 * @param bool         $wp_error         @see wp_update_post(). Optional. Whether to return a WP_Error on failure. Default false.
+	 * @param bool         $fire_after_hooks @see wp_update_post(). Optional. Whether to fire the after insert hooks. Default true.
+	 * @return int|WP_Error The post ID on success. The value 0 or WP_Error on failure.
+	 */
+	public static function update_post_without_modified_date_with_review( array|object $postarr, bool $wp_error = false, bool $fire_after_hooks = true ) {
+		// Validate $postarr['ID'] is set.
+		if ( ! isset( $postarr['ID'] ) ) {
```
naxoc: If `$postarr` is an object then this will fail. Can we add a check for either?

naxoc (commented):
This is great! Thank you. I left a small comment

src/Logic/Posts.php
iuravic: @naxoc , ah, of course, sorry! And thank you :D Now fixed in [c2bee8e](https://github.com/Automattic/newspack-migration-tools/pull/131/commits/c2bee8eed1d318e08b8a73aee894b7521feecb8e)

naxoc (approved):
Looks good! 👍

ronchambers:
@iuravic - I added some tweaks to a PR if you want to check them out:  https://github.com/Automattic/newspack-migration-tools/pull/132

PR #132, opened 2025-11-26
Update post with pre-revision

ronchambers:
Tweaks for PR https://github.com/Automattic/newspack-migration-tools/pull/131

* Explains that wp_update_post will still do a revision after the update, but this function also saves a revision pre-update.  WordPress will then have 2 revisions in the wp-admin > post editor.  A pre-update revision, and a revision that matches the updated content too.  WordPress expects the latest revision to be the same as the latest content.
* Explains why there is no default array.
* Fixes function name `...with_review` to `..._with_pre_revision` for most descriptive, longest name possible ;-)

### DIFF
--- a/src/Logic/Posts.php
+++ b/src/Logic/Posts.php
@@ -792,9 +792,11 @@ SQL;
 
 	/**
 	 * Shorthand for updating a post without touching the post_modified and post_modified_gmt fields,
-	 * but it saves a revision before the update.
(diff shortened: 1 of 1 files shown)


iuravic (approved):

iuravic:
Thank you @ronchambers this is great! I'm proceeding to merge this now, and then merging the #131 .

PR #134, opened 2025-11-28
Add an "original value store"

naxoc:
See #133

This adds a more generic and generally useful "value store" intended for migration data we want to keep for a while, like some IDs, original bylines, etc.

When reviewing, please start with the README and the two .md files added in this PR. I want to make sure this is well documented for both us humans, and the LLMs.
Then move on to the code – there is a lot of it, but most is command stuff.


kariae (approved):
@naxoc, this is top-notch. Thank you so much! 🎉 

One nit. We can skip adding `_post`, `_term`, and `_user` to the meta name, as they're stored in different places. It will make the meta name smaller and easier to read. Otherwise, this is ready to be merged!

naxoc:
> We can skip adding _post, _term, and _user to the meta name

Excellent point :) Not sure what I was thinking. I simplified it to just use `key_for()`.

PR #135, opened 2025-12-01
Ghost CMS use Guest Contributors instead of Guest Authors

iuravic:
Update to Ghost CMS migrator to code to use Guest Contributor role instead of Guest Authors.


src/Logic/GhostCMSHelper.php
```diff
@@ -345,85 +338,104 @@ private function get_or_import_url( string $path, string $title, string $caption
...
+				'number'         => 1,
 			)
 		);
+		$users         = $user_query->get_results();
+		if ( ! empty( $users ) ) {
+			$wp_user_found = $users[0];
+		}
 
-		foreach ( $user_query->get_results() as $wp_user ) {
+		// If we found an existing WP User with editing capabilities OR Guest Contributor role, use it.
```
ronchambers: Is this an extra check to make sure the edit capability actually exists?  Since the `role__in` user query above gets users by role, this still doesn't verify if the individual user had their `edit` capabilities removed by some other process, so doing this extra check for the specific capability would catch this case?  Is that right?  Nice!

It seems being extra cautious here make sense since the user query is just grabbing the first person by display name (where their could be multiple users with the same display name), so making sure that person has actual `edit` caps make is safer...but still it could be the wrong user if multiple users with same display name.  

I wonder if it makes more sense to not match on existing users and only match on users the migrator imported with `$unique_identifier`.  After the migration the publisher could delete any Guest Contributor and re-assign posts to the correct "pre-existing" user by hand if needed.  Just idea.  

**Aha...I see in my old code:** 
`// Return the first user found.`
Maybe this was a bad idea to begin with?  Maybe we shouldn't mess with existing users and only use `$unique_identifier` users?

ronchambers (commented):
Looks great!  The only issue I see is related to how I originally wrote the existing user matching so maybe now that we're using `$unique_identifier` we shouldn't do any "pre-existing" user matching lookups like the original code was doing?

src/Logic/GhostCMSHelper.php
iuravic: Thank you for this check @ronchambers ! I haven't looked into this preexisting cap check at all actually, just kept it, so I do appreciate your thoughts! Here's a couple of thoughts after checking it.

Perhaps it's because of my specific migration scenario right now, where I'm migrating and merging three different sites (one is Ghost, but other two are non-Ghost) into a single new Newspack site, that it seems particularly useful to reuse existing users, even without them having the unique ID (other migrators might not assign the same UIDs). And so I've now updated this logic to:
- remove the cap check -- IMHO this check was a nice thing with previous roles, but no longer needed
- and now doing the fetching of existing users in a more refined order:
    1. first by unique ID
    2. then by email -- email seems to me the second most unique identifier, because it can eliminate duplicating users like "R. Chambers" and "Ron Chambers"; if the email is the same, it's the same person
    3. then lastly searching by display name -- with clear warnings being displayed in such a scenario, so we do have insight and are able to double check if that was the intended John Doe from the logs

Perhaps this updated logic is probably fine now. I didn't refine every detail since I've got 2 more sites to migrate, but if you could give it another look and catch if anything really important is missing, I'd really appreciate it!

These things are handled in:
- 15c116429c652e5e8b0f3c1a3551374cf2c7042a .

---

Then I also did the importing of remaining user data/meta:
- 3fd40ab02372601bfdead736167811150e97bd57
- 080f24452944490978c7f37fdd115957e991a7d8

---

Lastly I fixed and updated tests:
- fixed the broken test by adding the SLA plugin as test dependency
- update the test to assert Guest Contributor role and check migrated additional user data/meta

in:
- 9c873eceb5ac594dfc13acc6e914856af9b734fc
- ef0c8410a4787777888e58bcdc0d460d29886f76

ronchambers (approved):
Looks awesome!  Approved 🚀

Issue #138, opened 2025-12-03
Download function during sideload will always add double extension: .jpg.jpg


ronchambers:
If `$desired_filename` is not set, then imported image will have duplicated extension: `.jpg.jpg`, `.png.png`, etc.

Related commits:

* https://github.com/Automattic/newspack-migration-tools/pull/115/commits/1fe05faee754d70a70d4ddb68084dc7dd2c83ec0 (check was added)
* https://github.com/Automattic/newspack-migration-tools/pull/115/commits/6c92a2b09b50570e20638d201f6630516f2ef488 (desired filename passed to function)

The new check always be false: `array_search( pathinfo( $path, PATHINFO_EXTENSION ), wp_get_mime_types() )`.

Based on how `wp_get_mime_types` array is structured, the check should probably be one of the following:

* Array key exists: `array_key_exists( pathinfo( $path, PATHINFO_EXTENSION ), wp_get_mime_types() )` - but this doesn't match on `jpg|jpeg|jpe`.

* Mime match: `array_search( mime_content_type( $tmpfname ), wp_get_mime_types() )`

I'm not quite sure why that check is there...maybe to make sure the file extension matches the mime and if not, add the correct extension?

For now, setting a `$desired_filename` before calling the import will bypass the check.

ronchambers:
@jorostoyanov - can you give this a look?  Thanks!

jorostoyanov:
Hmm, it makes perfect sense that the result will be always `false` for `array_search( pathinfo( $path, PATHINFO_EXTENSION ), wp_get_mime_types() )`. I will open a PR with a fix. Thank you for noticing that, @ronchambers!

> I'm not quite sure why that check is there...maybe to make sure the file extension matches the mime and if not, add the correct extension?

Exactly!

jorostoyanov:
Hey @ronchambers 👋 

I submitted a new [PR](https://github.com/Automattic/newspack-migration-tools/pull/141) that fixes this issue. In addition, I added two tests to validate the fix.

Can you have a look?

PR #141, opened 2025-12-09
Attachments: Prevent file extension duplication on download

jorostoyanov:
## Summary

This PR fixes an issue where you get a duplicate file extension when downloading a file with the Attachments Helper.

## Related Issue

https://github.com/Automattic/newspack-migration-tools/issues/138


ronchambers:
Thanks @jorostoyanov - I ran some local tests, but I'm not sure if I'm thinking about this correctly. 

I see the code is verifying the `$path` has an extension (`$file_has_extension`), but when it's checking `$file_extension_is_supported` this is just checking if the extension string is allowed, but it's not checking if the actual file's binary mime type matches the extension or if the file binary mime is allowed - which was the intent in the issue: _to make sure the file extension matches the mime and if not, add the correct extension_.

Can you help me understand what's intended for all cases (that I could think of)?  

**Extension exists and is an allowed extension:**
When `$path` has allowed extension (`.jpg`), with matching allowed binary mime (`image/jpeg`): do not modify filename.
When `$path` has allowed extension (`.jpg`), with different allowed binary mime (`image/png`): append correct extension (`.png`).
When `$path` has allowed extension (`.jpg`), with different unallowed binary mime (`application/x-msdownload`): ??
When `$path` has allowed extension (`.jpg`), with unknown binary mime (`unknown`): ??

**Extension exists but is not an allowed extension:**
When `$path` has unallowed extension (`.exe`), with matching unallowed binary mime (`application/x-msdownload`): ??
When `$path` has unallowed extension (`.exe`), with different allowed binary mime (`image/png`): append correct extension (`.png`).
When `$path` has unallowed extension (`.exe`), with different unallowed binary mime (`application/x-shockwave-flash`): ??
When `$path` has unallowed extension (`.exe`), with unknown binary mime (`unknown`): ??

**Unknown extension:**
When `$path` has unknown extension (`.unknown`), with allowed binary mime (`image/png`): append correct extension (`.png`).
When `$path` has unknown extension (`.unknown`), with non allowed binary mime (`application/x-msdownload`): ??
When `$path` has unknown extension (`.unknown`), with unknown binary mime (`unknown`): ??

**Missing extension:**
When `$path` has no extension, with allowed binary mime (`image/png`): append correct extension (`.png`).
When `$path` has no extension, with unallowed binary mime (`application/x-msdownload`): ??
When `$path` has no extension, with unknown binary mime (`unknown`): ??

Or should these "??" cases just be passed along to the `media_handle_sideload` and let that function deal with these cases?

Thanks!

kariae:
@ronchambers , I think we can skip checking the mimetype as this is handled by `media_handle_sideload()` that calls `wp_check_filetype_and_ext`.

Issue #79, opened 2025-06-18
Remove temporary folders after doing tests.


ronchambers:
https://github.com/Automattic/newspack-migration-tools/blob/0be65d46c25db8b91492c08394bcaa82e4380c03/tests/Util/CsvWriterTest.php#L19-L23

I'm using Newspack Docker on Mac OS with zsh shell.  When I run `phpunit` the CSV tests will leave behind folders like the following:

```
drwxr-xr-x  2 root     root         64 Jun 18 02:15 175021294968522155e6f86/
drwxr-xr-x  2 root     root         64 Jun 18 02:15 175021294968522155e7cdf/
drwxr-xr-x  2 root     root         64 Jun 18 02:15 175021294968522155e80d0/
drwxr-xr-x  2 root     root         64 Jun 18 02:15 175021294968522155e8439/
drwxr-xr-x  2 root     root         64 Jun 18 02:19 17502131716852223365367/
drwxr-xr-x  2 root     root         64 Jun 18 02:19 17502131716852223365f32/
drwxr-xr-x  2 root     root         64 Jun 18 02:19 175021317168522233662ec/
drwxr-xr-x  2 root     root         64 Jun 18 02:19 17502131716852223366605/
```

These are written in the NMT folder itself, not the `/tmp` folder.  My environment var for `echo $TEMPDIR` is blank.

Let me know if you want me to investigate further.  It would be nice to have these removed after tests are run :-)

jorostoyanov:
@ronchambers 👋 

Fixed here — https://github.com/Automattic/newspack-migration-tools/pull/142

Can you have a look?

ronchambers:
Closed per PR above.

PR #142, opened 2025-12-09
Remove temp dir after tests complete

jorostoyanov:
## Summary

Removes the temporary directory being created when the CSVWriter tests are being executed.

## Related Issue

https://github.com/Automattic/newspack-migration-tools/issues/79

### DIFF
--- a/src/Util/CsvWriter.php
+++ b/src/Util/CsvWriter.php
@@ -27,7 +27,7 @@ class CsvWriter {
 		private string $filename
 	) {
 		// phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_fopen
-		$this->file_pointer = fopen( getcwd() . '/' . $this->filename, 'a+' );
+		$this->file_pointer = fopen( $this->filename, 'a+' );
 
 		if ( false === $this->file_pointer ) {
 			// phpcs:ignore WordPress.Security.EscapeOutput.ExceptionNotEscaped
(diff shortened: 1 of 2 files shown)


ronchambers:
I stepped through the code and noticed the primary issue is actually above in the setup.

The problem is this line:

`$random_dir_name = wp_unique_filename( $temp_dir, uniqid( time() ) );`

It's generating a unique name for use within the `$temp_dir`, but the next line that actually creates the folder is not using the temp dir, it's just creating it in the current directory:

`wp_mkdir_p( $random_dir_name );`

My suggestion is to create the folder in the $temp_dir since that's the folder `wp_unique_filename` used.

Do this by prepending `$temp_dir` here:

```
$random_dir_name = $temp_dir . wp_unique_filename( $temp_dir, uniqid( time() ) );
```

If you do that, then I don't really care about left-over files in my `/tmp/` path, but yes, adding `rmdir` to the teardown as you did in this PR is nice :-)

Now...this will solve the issue with temp vs local in the tests, but it will then bring up another problem.  The constructor in the CsvWriter is hard coded to use the local working folder.

https://github.com/Automattic/newspack-migration-tools/blob/745b70cb44a2b92d6003e8656dd45251b289fe11/src/Util/CsvWriter.php#L30

My suggestion is to change the fopen line to this:

`$this->file_pointer = fopen( $this->filename, 'a+' );`

This way relative paths will still work, but absolute paths like `/tmp/1234` will work too.

ronchambers (changes requested):
Requested changes above.

jorostoyanov:
Hey @ronchambers 👋 

Thanks for the feedback!

It should be fixed now.

ronchambers (approved):

ronchambers:
@jorostoyanov - Approved, thanks!

PR #143, opened 2025-12-12
Optional command argument for path to data in JSON file

iuravic:
Turns out The Philly Download has a slightly different path to posts (and other data) in their Ghost export JSON file:
- our migrator expected to find posts at `.db[0].data.posts`
- but this one has posts at just `.data.posts`

This PR adds an optional `--json-data-path` param, which takes standard `jq` JSON path notation to posts and other data:
- default value (to posts and other data) remains `.db[0].data` if nothing is provided
- but I was able to import The Philly Download using just `--json-data-path=".data"`

The new argument `--json-data-path` also has examples in command argument description on how one can use `jq` from CLI to check and determine the path.

### DIFF
--- a/docs/GhostCMS.md
+++ b/docs/GhostCMS.md
@@ -65,7 +65,7 @@ Standard `jq`-style path notation to node in JSON where posts (and other objects
 
 Default value path (to posts and other data) is `.db[0].data`.
 To test for path, you can use `jq` from CLI like:
-- count posts: `jq '.db[0].data.posts' | length export.json`
+- count posts: `jq '.db[0].data.posts | length' export.json`
 - list posts:  `jq '.db[0].data.posts' export.json`
 ```
 
--- a/docs/GhostCMS.md
+++ b/docs/GhostCMS.md
@@ -59,6 +59,14 @@ Optional arguments:
 ```
 --created-after=<created-after>
 Datetime cut-off to only import posts AFTER this date. (Must be parseable by strtotime).
+
+--json-data-path=<json-data-path>
+Standard `jq`-style path notation to node in JSON where posts (and other objects) are stored (e.g., --json-data-path=".db[0].data" or --json-data-path=".data").
+
+Default value path (to posts and other data) is `.db[0].data`.
+To test for path, you can use `jq` from CLI like:
+- count posts: `jq '.db[0].data.posts' | length export.json`
+- list posts:  `jq '.db[0].data.posts' export.json`
 ```
 
 ### Step 4: Run a test
@@ -75,7 +83,7 @@ For testing, you can use these test values (with the included `json` test file):
 
 Command (_be sure to replace your values_):
 ```
-wp newspack-migration-tools ghostcms-import --default-user-id=<default-user-id> --ghost-url=<ghost-url> --json-file=<json-file> [--created-after=<created-after>]
+wp newspack-migration-tools ghostcms-import --default-user-id=<default-user-id> --ghost-url=<ghost-url> --json-file=<json-file> [--created-after=<created-after>] [--json-data-path=<json-data-path>]
 ```
 
 If the migrator command is stopped mid-migration, it is OK to simply re-run the command.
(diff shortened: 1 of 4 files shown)


ronchambers:
Nice!  I wonder if this new format could be the case if a publisher only has one database at ghost so there is no "db" array?  Possibly:

JSON with one db:
```
{
    "data": {
```

JSON with multiple dbs? (Or multiple dbs, but only one exported).
```
{
  "db": [
    {
      "data": {
```

Maybe we need to put something in the docs that says "our default is to import the first database found in the JSON" and then we change the code to test if "db[] array" exists, if yes then we use "db[0].data", but if not then we just use ".data".  This would allow us to not have to specific the path, it would be chosen by the code.

But then again, if the JSON had multiple "dbs", the the new data path parameter would allow us to choose a different db `.db[X].data` :-)

Maybe we do a mix of both:
* Put a notice that the importer will use the first db in the JSON.
* Change the code to automatically choose ".data" if at root, if not, then choose "db[0].data".
* Keep the new `--json-data-path` but change the wording to say that we will automatically import the first db in the JSON, but if a different DB or path is needed then use this new parameter.

Totally up to you.  

I tested the PR as-is and it worked locally.

I would recommend these two changes though:
* In the examples, length should be inside the quotes: `jq '.db[0].data.posts | length' export.json`
* Add tests for the new function `get_json_data_from_path`

iuravic:
Thank you @ronchambers , now updated the [demo length jq command](https://github.com/Automattic/newspack-migration-tools/pull/143/commits/a27be98ee196903ada61c15baab16058b42c7204) and [added tests](https://github.com/Automattic/newspack-migration-tools/pull/143/commits/13622f1ee84be0306a6a6261ba28d49887655660). I also [removed an unnecessary regex](https://github.com/Automattic/newspack-migration-tools/pull/143/commits/5eb1eeabb5ba80de28d15cd916f10d6aa9b0502d) and replaced it with a simple explode.

> change the wording to say that we will automatically import the first db in the JSON, but if a different DB or path is needed then use this new parameter

As for this part, it seems to me that it should already be clear by this part of the description:

> 'Default value (path to posts) is `.db[0].data`.'

meaning it will "automatically import the first DB". As for hard-coding any other paths, I'd recommend we avoid it for now -- seems a bit too complex for such a simple operation. Plus we're encountering this new path for the first time, it might not be repeated. So perhaps we can leave it be for now, see if more logic is needed in the future. But as for better and more verbose argument description, I'm always for that. Just for this one it kind of seems informative enough as it is.

ronchambers:
Nice!  I'll test locally in a bit.

ronchambers (approved):
@iuravic this is awesome!  I'll approve as-is....but I'll present two options if you want to merge them in to this PR - up to you!

Option 1 ( [PR](https://github.com/Automattic/newspack-migration-tools/pull/144) ):

- Removes the `$json` property from the object now that `$this->data` is new class-wide property.
- Updates `get_json_data_from_path` with a secondary parameter for the local `$json` value so that `set_json` is not needed.  
- Also makes `get_json_data_from_path` public for easier testing.

Option 2 ( [PR](https://github.com/Automattic/newspack-migration-tools/pull/145) ):

- If option 1 isn't preferred, this this small tweak just fixes a bug if the json file is malformed.

Thanks!

iuravic:
@ronchambers , thanks so much for your thoughtful reviews, I love them! And as always I really appreciate your thoughts :high-five

I immediately accepted and merged https://github.com/Automattic/newspack-migration-tools/pull/145 , with that extra check on the JSON object.

And I also merged the first PR https://github.com/Automattic/newspack-migration-tools/pull/144 too with these updates:
- removed `$this->json property` -- thanks for clearing up this even further that
- removed `set_json`
- now passing `$json` as argument to method

The only thing I reverted was making the method public. When I started this PR, I too was originally thought about making it public, but kind of intentionally settled on private. Here was my reasoning, sharing as a complementary and additional view:
- there doesn't seem to be any need for it to be exposed as public
- what I personally appreciate as a general style, is the ability to look at class for the first time (overview of classes in any IDE or even GH's right side view), and glance through its public methods, and just from their naming being able to quickly understand what the class is doing as a logic entity. Keeping that set minimal might be said to make the functionality more compact
- additionally, I've grown accustomed of unit testing private methods with PHP reflection, so this already kind of seemed comfortable and clean

Now updated this branch and ready for a re-review.

ronchambers (approved):
Approved!  Your notes about public/private make total sense :-) Thanks!

PR #148, opened 2026-01-06
Fix: Prevent double file extensions when importing attachments

kariae:
### Description

This PR fixes a bug where files with valid extensions were incorrectly getting a duplicate extension appended during import (e.g., `image.jpg` becoming `image.jpg.jpg`).

### The Bug

In the `download_file()` method, the code was checking whether a file had a valid extension using:

```php
} elseif (
    ! pathinfo( $path, PATHINFO_EXTENSION )
    || ( false === array_search( pathinfo( $path, PATHINFO_EXTENSION ), wp_get_mime_types() ) )
) {
    // Add extension based on mime type...
    $file_array['name'] .= '.' . $probably_extension;
}
```

The problem is that `wp_get_mime_types()` returns an array structured like:

```php
[
    'jpg|jpeg|jpe' => 'image/jpeg',
    'png'          => 'image/png',
    // ...
]
```

The `array_search()` function searches in the **values** (the mime types), not the **keys** (the extensions). So `array_search('jpg', wp_get_mime_types())` would search for `'jpg'` in `['image/jpeg', 'image/png', ...]`, which **always returns `false`** because `'jpg'` never matches `'image/jpeg'`.

This caused the condition to be true for every file, even those with valid extensions, resulting in an extra extension being appended.

### The Fix

Extracted the extension validation logic into a new private method `has_valid_extension()` that uses `wp_check_filetype()`, which is the proper WordPress function for validating file extensions:

```php
private static function has_valid_extension( string $path ): bool {
    $extension = pathinfo( $path, PATHINFO_EXTENSION );

    if ( empty( $extension ) ) {
        return false;
    }

    // Use wp_check_filetype which properly checks extensions against allowed mime types.
    $filetype = wp_check_filetype( '.' . $extension );

    return ! empty( $filetype['ext'] );
}
```

### Testing

Two new unit tests were added to `tests/Logic/TestAttachmentHelper.php`:

1. **`test_download_file_does_not_create_double_extension`** - Regression test that verifies files with valid extensions (like `.jpg`) don't end up with double extensions (like `.jpg.jpg`)

2. **`test_download_file_adds_extension_when_missing`** - Verifies that files without extensions still correctly get an extension added based on mime type detection

### How to Test

```bash
# Run the new tests
./vendor/bin/phpunit --filter TestAttachmentHelper

# Or run just the regression test
./vendor/bin/phpunit --filter test_download_file_does_not_create_double_extension
```

### Checklist

- [x] Bug fix with regression test
- [x] No breaking changes
- [x] Tests pass locally

### DIFF
--- a/src/Logic/Attachments.php
+++ b/src/Logic/Attachments.php
@@ -306,10 +306,7 @@ class Attachments {
 
 		if ( ! empty( $desired_filename ) ) {
 			$file_array['name'] = $desired_filename;
-		} elseif (
-			! pathinfo( $path, PATHINFO_EXTENSION )
-			|| ( false === array_search( pathinfo( $path, PATHINFO_EXTENSION ), wp_get_mime_types() ) )
-		) {
+		} elseif ( ! self::has_valid_extension( $path ) ) {
 			// If the path does not have a file extension, let's try to find one for it.
 			// Without the extension, the upload will fail because WP will not allow that "file type".
 			$mimetype           = mime_content_type( $tmpfname );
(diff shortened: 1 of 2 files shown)


kariae:
Closing in favor of #141

PR #149, opened 2026-01-08
Fix/prevent file extension duplicate Ron 2

ronchambers:
todo:
- phpcs
- don't do http requests (if possible?) or use wp_safe_remote hook...

## Changes to Helper

- Updated docblock.
- Return early if desired filename is set.
- Verify extension against all know extensions `wp_get_mime_types` (not just "allowed" extensions, otherwise unallowed files become ".swf.swf".
- Minor refactoring.
  - Not copying to temp file unless checking that local file exists first.
  - If temp file size is empty, then remove then unlink the temp file.
- For clarity: even though many of the WordPress functions use the word "mime" in their function name, our helper is not checking the binary mime type for each download - we're only verifying the file extension.  Only in the cases where  the extension check returns `false` do we actually check the mime with `mime_content_type`.

## Fixtures

- Renamed "koi" (no extension) to `image-jpeg.jpeg`, then duplicated image file to different filenames for tests.
- Copied other files (variety of extensions and mimes) from [WP Core test files](https://github.com/WordPress/wordpress-develop/tree/ac6e7c904a39fabaa0f3d3770554f05373552db4/tests).
- Created `html` and `swf` files by hand as noted in the test file.
- These files give us a variety of extensions and mimes for extensive test coverage.

## Tests

Each test is defined by multiple properties, but the most important are:

```
'path' => path to file or URL
'downloaded-file-name' => the filename returned by download_file.
'sideloaded-file-name' => the subsequent filename after media_handle_sideload.
'sideloaded-wp-core'   => unmodified filename for default wp core sideload test comparison.
```

The tests cover many cases, such as `image.jpg`, wrong extensions `jpeg-image.png`, missing extensions `image`, unknown extensions `file.unknown`, unallowed extensions `flash.swf`, and "hacked files" with allowed extension but unallowed mime like `flash.mov`

Also URLs with `?querystring` are tested too.

Tests use "wrap" functions so that all files stored in `/tmp/` and as `wp-uploads` are removed during `teardown`.


kariae:
@ronchambers @jorostoyanov can we handle the last few meters to merge this, so that it's included for future migrations. Thank you!

jorostoyanov:
@kariae Sure thing! I have some tests running, and will be ready very soon

jorostoyanov (approved):
LGTM after running tests! 👍 
Good job, @ronchambers

PR #153, opened 2026-01-29
Ghost migrator addons

iuravic:
### Migrate post `custom_excerpt` field to Newspack subititle
Added in [21aaac7](https://github.com/Automattic/newspack-migration-tools/pull/153/commits/21aaac7cd02cdb9e6158db41b97b77f84ebbe2a5).

Example post
- live Ghost post with subtitle -- https://www.theicegarden.com/seattle-torrent-break-record-arrive-with-purpose-in-their-first-home-game/
- original Ghost export record for this post: [ghost_id_692e2b5122ad6e00017d5407.json](https://github.com/user-attachments/files/24936648/ghost_id_692e2b5122ad6e00017d5407.json)

### Warn if there are additional custom post `visibility` values which are not being migrated
Added in [418aa86](https://github.com/Automattic/newspack-migration-tools/pull/153/commits/418aa86c5beb3bda61b7451e9ed134700a0c9da0).

I just noticed that a bunch of posts were not migrated for one Publisher, and was not aware of it. The logs do show `Skip JSON post (review by hand -skips.log): not_public` but that is not entirely true, as that Ghost site has multiple public statuses, not just `public`, but also `member`, `paid`, `tiers`.

This now:
- scans all the existing `visibility` values in posts before importing the posts
- if there are any additional statues which are not being migrated, it outputs a warning and prompts the user whether they wish to proceed
- uses a different skip status, instead of `not_public` it is now labeled `visibility_is_different`.

Example, there are multiple other values besides the default `public`, and the user is warned:
Screenshot 2026-01-29 at 4 42 00 PM

Example, `--visibility-csv=public,paid,members` was provided, but there's also `tiers`, so the user is warned and prompted whether they wish to proceed:
Screenshot 2026-01-29 at 4 41 23 PM

No warning/confirmation is shown if `public` is the only visibility which exists.

### DIFF
--- a/src/Command/GhostCMSMigrator.php
+++ b/src/Command/GhostCMSMigrator.php
@@ -60,7 +60,9 @@ class GhostCMSMigrator implements WpCliCommandInterface {
 							'type'        => 'assoc',
 							'name'        => 'visibility-csv',
 							'description' => 'Comma separated list of post visibility (i.e. post status) values to import. ' .
-											'This command will scan for existing visibilities in the JSON data before importing any posts, and warn you if there are any multiple values besides the default `public`. ' .
+											'This command will scan for existing visibilities in ALL posts in the JSON data before applying any filters (like --created-after). ' .
+											'You will be warned if there are multiple visibility values besides the default `public`. ' .
+											'Note: The visibility scan in this command reports on the entire dataset, which may include visibility values not present in the filtered date range. ' .
 											'E.g. `--visibility-csv=public,members,paid,tiers`.',
 							'optional'    => true,
 							'repeating'   => false,
--- a/src/Command/GhostCMSMigrator.php
+++ b/src/Command/GhostCMSMigrator.php
@@ -58,6 +58,15 @@ class GhostCMSMigrator implements WpCliCommandInterface {
 						// optional:
 						array(
 							'type'        => 'assoc',
+							'name'        => 'visibility-csv',
+							'description' => 'Comma separated list of post visibility (i.e. post status) values to import. ' .
+											'This command will scan for existing visibilities in the JSON data before importing any posts, and warn you if there are any multiple values besides the default `public`. ' .
+											'E.g. `--visibility-csv=public,members,paid,tiers`.',
+							'optional'    => true,
+							'repeating'   => false,
+						),
+						array(
+							'type'        => 'assoc',
 							'name'        => 'created-after',
 							'description' => 'Datetime cut-off to only import posts AFTER this date. (Must be parseable by strtotime).',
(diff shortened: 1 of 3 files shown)


src/Command/GhostCMSMigrator.php
ronchambers, line 1: Future suggestion: put a note in the Ghost doc about this new argument and mention the visibility check scans all the posts before the `--created-after` argument is applied, which means the visibility checker might report a visibility that is not actually in the set of posts that will be imported with `--created-after`.

src/Logic/GhostCMSHelper.php
ronchambers, line 264: Future idea: possibly store the original visibility value from the JSON into a postmeta for future reference if needed, like `_newspack_ghostcms_visibility`

src/Logic/GhostCMSHelper.php
```diff
@@ -1015,4 +1042,21 @@ public function replace_audio_embeds( string $content ): string {
...
+
+	/**
+	 * Get all visibility values from JSON data.
+	 *
+	 * @param object $data JSON data.
+	 * @return array Visibility values.
+	 */
+	private function get_visibility_values( object $data ): array {
+		$visibilities = [];
+		foreach ( $data->posts as $json_post ) {
```
ronchambers, line 1054: Future: since other JSON nodes have `visibility`, this function could be more generalized like `foreach ( $data as $item )` so it's not just for `posts`.  The the calling code would pass in the posts node like the following `$visibilities_existing = $this->get_visibility_values( $this->data->posts );`

ronchambers (approved):
@iuravic - this is awesome!  Clean, precise, works exactly as expected, easy to review 😄  Thank you!

It is APPROVED.  Feel free to ignore my comments.  They were just some "future/ideas" I noted while doing the review so I shared them.

src/Command/GhostCMSMigrator.php
iuravic: Good point, @ronchambers . Added in [9eda6dc](https://github.com/Automattic/newspack-migration-tools/pull/153/commits/9eda6dcd44612ac0070d615bf4b3347a8ed0d131):
- updated command argument with this point,
- added the new argument to Ghost migrator's README with description

src/Logic/GhostCMSHelper.php
iuravic: Thanks for the idea, this did cross my mind but it seems that this would duplicate the meta field so I thought it wasn't necessary for this scenario.

iuravic:
Thanks @ronchambers 🙏  I added the missing readme entry, and extended the argument description, and am about to merge now.

PR #158, opened 2026-02-09
Deprecate the duplicate get_attachment_by_filename in favor of get_attachment_id_by_filename

iuravic:
The newer method has proper LIKE escaping, strict return type (int), guaranteed non-null return (0 instead of null), and we should remove the older duplicate (I additionally updated the duplicate's docblock return type description because it was wrong).


kariae (approved):
Let's go!

PR #162, opened 2026-02-13
feat: init commit of CommandSupervisor

eddiesshop:
Summary
--
  Adds a new CommandSupervisor WP-CLI command to newspack-migration-tools that wraps and supervises the execution of other WP-CLI commands. It automatically restarts commands that fail due to crashes, timeouts, or OOM kills -- a common occurrence during long-running content migrations.

  - Automatic retry on failure with configurable total and consecutive failure limits
  - Restart-on-success mode for batch-processing commands that need multiple passes
  - Exit-code-based completion signaling (NMT::EXIT_DONE) -- supervised commands call WP_CLI::halt( NMT::EXIT_DONE ) to tell the supervisor they're finished. String-based --completion-criteria is available as a fallback for commands that haven't adopted the exit code convention.
  - Memory optimization via automatic `--skip-themes` and `--skip-plugins` flag injection
  - Default plugin whitelist -- Co-Authors Plus, Simple Local Avatars, Yoast, and Newspack are kept active by default, along with any plugin that depends on automattic/newspack-migration-tools (detected dynamically via composer.json)
  - Logging to both CLI and file via MultiLog
  - Optional email notification when supervision ends

  Why
--
  Migration scripts often run for hours and can fail unpredictably due to memory limits or migration issues. Today, engineers have to manually monitor and restart these scripts. CommandSupervisor automates this, making unattended migration runs possible.

  For batch-processing migrations (e.g., "process 500 posts per run"), the --restart-on-success mode lets the supervisor keep re-running the command until it signals completion via `NMT::EXIT_DONE`, without human intervention.

  Usage
--
  Basic: auto-restart on failure

```bash
wp newspack-content-migrator command-supervisor \
    --command="newspack-content-migrator some-migration --batch-size=500"
```

  Runs the migration command. If it crashes, waits 5 seconds and retries. Stops after 3 total failures or 2 consecutive failures.

  Batch processing with EXIT_DONE (preferred)

```bash
wp newspack-content-migrator command-supervisor \
    --command="newspack-content-migrator migrate-posts --batch-size=200" \
    --restart-on-success
```
  Keeps re-running the command after each successful exit (exit code 0). The supervised command signals completion by calling WP_CLI::halt( NMT::EXIT_DONE ) (exit code 2) when there's no more work. No additional flags needed.

  Batch processing with completion criteria (fallback)

```bash
wp newspack-content-migrator command-supervisor \
    --command="newspack-content-migrator migrate-posts --batch-size=200" \
    --restart-on-success \
    --completion-criteria="No more posts to process"
```
  For commands that haven't adopted NMT::EXIT_DONE, the supervisor can scan output for a completion string instead.

  Batch processing with a fixed number of passes

```bash
wp newspack-content-migrator command-supervisor \
    --command="newspack-content-migrator migrate-posts --batch-size=200" \
    --restart-on-success \
    --max-success-retries=20
```
  Re-runs the command up to 20 successful times, then stops. Useful when you know roughly how many passes are needed.

  With email notification and custom retry settings

```bash
wp newspack-content-migrator command-supervisor \
    --command="newspack-content-migrator heavy-migration" \
    --max-fail-retries=5 \
    --max-consecutive-fail-retries=3 \
    --retry-delay=30 \
    --notify-email=engineer@example.com
```

  Allows up to 5 total failures (3 consecutive) with a 30-second delay between retries. Sends an email when supervision ends.

  Keeping specific plugins active

```bash
wp newspack-content-migrator command-supervisor \
    --command="newspack-content-migrator migrate-woo-orders" \
    --active-plugins="woocommerce,woocommerce-subscriptions"
```

  By default, all themes and non-essential plugins are skipped for memory optimization. Co-Authors Plus, Simple Local Avatars, Yoast, and Newspack are kept active by default, along with any plugin that depends on newspack-migration-tools (detected via composer.json). Use --active-plugins to whitelist additional plugins the supervised command depends on.

  Parameters

| Parameter| Required | Default |Description |
|--------|--------|--------|--------|
| --command| Yes | -- | The WP-CLI command to supervise (without the wp prefix)|
| --max-fail-retries | No | 3 | Max total failed attempts before stopping |
| --max-consecutive-fail-retries | No | 2 | Max consecutive failures before stopping |
| --retry-delay | No | 5 | Seconds to wait between retries |
| --restart-on-success | No | -- | Also restart after successful exits (exit code 0). The command can signal completion via exit code 2 (NMT::EXIT_DONE). |
| --completion-criteria | No | -- | Fallback: string to look for in output that signals "done" (for commands that don't use NMT::EXIT_DONE) |
| --max-success-retries | No | 10 | Max successful runs (only when --restart-on-success is set without --completion-criteria) |
| --active-plugins | No | -- | Comma-separated additional plugin slugs to keep active (Co-Authors Plus, Simple Local Avatars, Yoast, and Newspack are kept active by default) |
| --notify-email | No | --| Email address to notify when supervision ends |

  Test plan
--
  - Unit tests added (38 tests, 84 assertions) covering:
    - Command registration and synopsis structure
    - Global flag injection (`--skip-themes`, `--skip-plugins`)
    - Plugin skip-list building with whitelist support
    - Default migration plugin whitelist (Co-Authors Plus, Simple Local Avatars, Yoast, Newspack)
    - Dynamic host plugin detection via composer.json
    - WP prefix stripping from command input
    - Email notification content for success and failure
    - `NMT::EXIT_DONE` constant definition
    - Supervision loop: `EXIT_DONE` stops loop with and without `--restart-on-success`
    - Supervision loop: exit 0 continues with `--restart-on-success`, stops without
    - Supervision loop: `EXIT_DONE` normalizes final exit code to 0 in notifications
    - Supervision loop: `--max-success-retries` stops the loop

### DIFF
--- a/src/Command/CommandSupervisor.php
+++ b/src/Command/CommandSupervisor.php
@@ -2,6 +2,7 @@
 
 namespace Newspack\MigrationTools\Command;
 
+use Newspack\MigrationTools\NMT;
 use Newspack\MigrationTools\Util\Log\MultiLog;
 use Psr\Log\LoggerInterface;
 use WP_CLI;
@@ -73,13 +74,13 @@ class CommandSupervisor implements WpCliCommandInterface {
 						[
 							'type'        => 'flag',
 							'name'        => 'restart-on-success',
-							'description' => 'Also restart the command after a successful exit (exit code 0). Useful for batch-processing commands that need to run multiple passes. Requires --completion-criteria or --max-success-retries.',
+							'description' => 'Also restart the command after a successful exit (exit code 0). The supervised command can signal completion by exiting with code 2 (EXIT_DONE). Optionally use --completion-criteria or --max-success-retries as alternative stop conditions.',
 							'optional'    => true,
 						],
 						[
 							'type'        => 'assoc',
 							'name'        => 'completion-criteria',
-							'description' => 'A string to look for in the command output that indicates processing is complete and the supervisor should stop restarting. When set, --max-success-retries is ignored.',
+							'description' => 'A string to look for in the command output that indicates processing is complete. Fallback for commands that cannot use exit code 2 (EXIT_DONE) to signal completion. When set, --max-success-retries is ignored.',
 							'optional'    => true,
 						],
 						[
@@ -92,7 +93,7 @@ class CommandSupervisor implements WpCliCommandInterface {
 						[
 							'type'        => 'assoc',
 							'name'        => 'active-plugins',
-							'description' => 'Comma-separated list of additional plugin slugs to keep active. Plugins that depend on newspack-migration-tools are automatically kept active. All other plugins are skipped via --skip-plugins for memory optimization.',
+							'description' => 'Comma-separated list of additional plugin slugs to keep active. Co-Authors Plus, Simple Local Avatars, Yoast, and Newspack are kept active by default, along with any plugin that depends on newspack-migration-tools. All other plugins are skipped via --skip-plugins for memory optimization.',
 							'optional'    => true,
 						],
 						[
@@ -154,9 +155,16 @@ class CommandSupervisor implements WpCliCommandInterface {
 
 			$this->logger->info( sprintf( '========== Attempt #%d ==========', $attempt ) );
 
-			$process         = $this->execute_command( $command );
-			$final_exit_code = $process->return_code;
-			$success         = ( 0 === $process->return_code );
+			$process = $this->execute_command( $command );
+			// If return code is 2 (EXIT_DONE) is received, the command has fully completed its task.
+			$done = ( NMT::EXIT_DONE === $process->return_code );
+			// Want to know if the command has succeeded or not. A return code of 1 (EXIT_ERROR) means an error occurred.
+			// A return code of 1 is ignored if EXIT_DONE is received.
+			$success = ( 0 === $process->return_code || $done );
+			// If EXIT_DONE is received, we convert the final exit code to 0, indicating success. Otherwise, fallback to whatever the final exit code is.
+			$final_exit_code = $done ? 0 : $process->return_code;
+			// Track the success/failure status for the current and previous attempts.
+			// This is used to determine consecutive successes or failures.
 			if ( null !== $operation_status_stack[0] ) {
 				$operation_status_stack[1] = $operation_status_stack[0];
 			}
@@ -169,6 +177,12 @@ class CommandSupervisor implements WpCliCommandInterface {
 			if ( $success ) {
 				++$total_success_count;
 				$consecutive_fail_count = 0;
+
+				if ( $done ) {
+					$this->logger->info( sprintf( 'Command signaled completion (exit code %d) on attempt #%d. Stopping.', NMT::EXIT_DONE, $attempt ) );
+					break;
+				}
+
 				$this->logger->info( sprintf( 'Command succeeded on attempt #%d (exit code 0).', $attempt ) );
 
 				if ( ! $restart_on_success ) {
@@ -251,7 +265,7 @@ class CommandSupervisor implements WpCliCommandInterface {
 	 * @param string $command WP-CLI command without the "wp" prefix.
 	 * @return object{stdout: string, stderr: string, return_code: int} Process result.
 	 */
-	private function execute_command( string $command ): object {
+	protected function execute_command( string $command ): object {
 		return WP_CLI::runcommand(
 			$command,
 			[
@@ -319,12 +333,13 @@ class CommandSupervisor implements WpCliCommandInterface {
 		// Build the whitelist: host plugin(s) + default active plugins + caller-specified plugins.
 		$whitelist = $this->get_host_plugin_slugs( $all_active );
 
-		// Allow default active plugins to be configured via constant or filter.
-		$default_active_plugins = [];
-		if ( defined( 'NMT_DEFAULT_ACTIVE_PLUGINS' ) ) {
-			$default_active_plugins = array_map( 'trim', explode( ',', NMT_DEFAULT_ACTIVE_PLUGINS ) );
-		}
-		$default_active_plugins = apply_filters( 'newspack_migration_tools_default_active_plugins', $default_active_plugins );
+		// Plugins commonly used during migrations that should remain active by default.
+		$default_active_plugins = [
+			'co-authors-plus',
+			'simple-local-avatars',
+			'wordpress-seo',
+			'newspack-plugin',
+		];
 		$whitelist              = array_merge( $whitelist, $default_active_plugins );
 
 		if ( ! empty( $active_plugins_arg ) ) {
--- a/src/Command/CommandSupervisor.php
+++ b/src/Command/CommandSupervisor.php
@@ -283,15 +283,11 @@ class CommandSupervisor implements WpCliCommandInterface {
 
 		$flags = [];
 
-		if ( $has_skip_themes ) {
-			$this->logger->info( '--skip-themes already present in command; skipping.' );
-		} else {
+		if ( ! $has_skip_themes ) {
 			$flags[] = '--skip-themes';
 		}
 
-		if ( $has_skip_plugins ) {
-			$this->logger->info( '--skip-plugins already present in command; skipping.' );
-		} else {
+		if ( ! $has_skip_plugins ) {
 			$skip_plugins_list = $this->get_plugins_to_skip( $active_plugins_arg );
 			if ( ! empty( $skip_plugins_list ) ) {
 				$flags[] = '--skip-plugins=' . implode( ',', $skip_plugins_list );
--- a/src/Command/CommandSupervisor.php
+++ b/src/Command/CommandSupervisor.php
@@ -316,8 +316,17 @@ class CommandSupervisor implements WpCliCommandInterface {
 	private function get_plugins_to_skip( string $active_plugins_arg ): array {
 		$all_active = get_option( 'active_plugins', [] );
 
-		// Build the whitelist: host plugin(s) + caller-specified plugins.
+		// Build the whitelist: host plugin(s) + default active plugins + caller-specified plugins.
 		$whitelist = $this->get_host_plugin_slugs( $all_active );
+
+		// Allow default active plugins to be configured via constant or filter.
+		$default_active_plugins = [];
+		if ( defined( 'NMT_DEFAULT_ACTIVE_PLUGINS' ) ) {
+			$default_active_plugins = array_map( 'trim', explode( ',', NMT_DEFAULT_ACTIVE_PLUGINS ) );
+		}
+		$default_active_plugins = apply_filters( 'newspack_migration_tools_default_active_plugins', $default_active_plugins );
+		$whitelist              = array_merge( $whitelist, $default_active_plugins );
+
 		if ( ! empty( $active_plugins_arg ) ) {
 			$whitelist = array_merge(
 				$whitelist,
--- a/src/Command/CommandSupervisor.php
+++ b/src/Command/CommandSupervisor.php
@@ -168,6 +168,7 @@ class CommandSupervisor implements WpCliCommandInterface {
 
 			if ( $success ) {
 				++$total_success_count;
+				$consecutive_fail_count = 0;
 				$this->logger->info( sprintf( 'Command succeeded on attempt #%d (exit code 0).', $attempt ) );
 
 				if ( ! $restart_on_success ) {
@@ -210,6 +211,11 @@ class CommandSupervisor implements WpCliCommandInterface {
 					$this->logger->warning( sprintf( 'Max fail retries (%d) reached. Stopping.', $max_fail_retries ) );
 					break;
 				}
+
+				if ( $consecutive_fail_count >= $max_consecutive_fail_retries ) {
+					$this->logger->warning( sprintf( 'Max consecutive fail retries (%d) reached. Stopping.', $max_consecutive_fail_retries ) );
+					break;
+				}
 			}
 
 			$this->logger->info( sprintf( 'Retrying in %d seconds...', $retry_delay ) );
@@ -228,9 +234,7 @@ class CommandSupervisor implements WpCliCommandInterface {
 			}
 		}
 
-		if ( $total_success_count >= $max_success_retries ) {
-			$this->logger->info( 'Command succeeded after max success retries.' );
-		} elseif ( $consecutive_fail_count >= $max_consecutive_fail_retries || $total_fail_count >= $max_fail_retries ) {
+		if ( $consecutive_fail_count >= $max_consecutive_fail_retries || $total_fail_count >= $max_fail_retries ) {
 			$this->logger->error( 'Command failed after max consecutive failures or max fail retries.' );
 			WP_CLI::halt( 1 );
 		}
--- a/src/Command/CommandSupervisor.php
+++ b/src/Command/CommandSupervisor.php
@@ -117,10 +117,10 @@ class CommandSupervisor implements WpCliCommandInterface {
 	 */
 	public function cmd_supervise( array $args, array $assoc_args ): void {
 		$command              = $assoc_args['command'];
-		$max_fail_retries     = intval( $assoc_args['max-fail-retries'] ?? 3 );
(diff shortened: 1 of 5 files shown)


src/Command/CommandSupervisor.php
```diff
@@ -0,0 +1,345 @@
...
+			++$attempt;
+
+			$this->logger->info( "========== Attempt #{$attempt} ==========" );
+
+			$process         = $this->execute_command( $command );
+			$final_exit_code = $process->return_code;
+			$success         = ( 0 === $process->return_code );
+
+			if ( ! empty( $process->stdout ) ) {
+				$this->logger->info( $process->stdout );
```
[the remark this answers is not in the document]
eddiesshop: This is incorrect. Using `all => true` will tell WP_CLI to return a `stdClass` object. See: https://github.com/wp-cli/wp-cli/blob/main/php/class-wp-cli.php#L1520-L1526

src/Command/CommandSupervisor.php
```diff
@@ -0,0 +1,345 @@
...
+						[
+							'type'        => 'assoc',
+							'name'        => 'command',
+							'description' => 'The WP-CLI command to supervise, without the "wp" prefix (e.g. "newspack-content-migrator some-command --batch-size=100").',
+							'optional'    => false,
+						],
+						[
+							'type'        => 'assoc',
+							'name'        => 'max-fail-retries',
+							'description' => 'Maximum number of consecutive or total failed attempts before giving up. Successful runs do not count toward this limit.',
```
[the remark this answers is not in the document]
eddiesshop: I made this flag two distinct flags: max _consecutive_ failure retries and max failure retries. If something fails consecutively, it likely means that the script should stop doing what it's doing. Having the flag gives the caller the option to increase that threshold.

Having max failure retries allows the caller to establish if there's an acceptable range of failures across the script's runtime.

src/Command/CommandSupervisor.php
```diff
@@ -0,0 +1,345 @@
...
+
+			$this->logger->info( "========== Attempt #{$attempt} ==========" );
+
+			$process         = $this->execute_command( $command );
+			$final_exit_code = $process->return_code;
+			$success         = ( 0 === $process->return_code );
+
+			if ( ! empty( $process->stdout ) ) {
+				$this->logger->info( $process->stdout );
+			}
```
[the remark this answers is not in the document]
eddiesshop: There are already two built-in mechanisms for this — no code changes needed in CommandSupervisor:

  1. NMT_LOG_DIR constant — define it in wp-config.php to point to a private directory:
  `define( 'NMT_LOG_DIR', WP_CONTENT_DIR . '/private-logs' );`
  2. newspack_migration_tools_log_dir filter — hook into it to return a private path:
  ```php
add_filter( 'newspack_migration_tools_log_dir', function() {
      $private_dir = WP_CONTENT_DIR . '/private-logs';
      if ( ! is_dir( $private_dir ) ) {
          wp_mkdir_p( $private_dir );
      }
      return $private_dir;
  });
```

  Both are checked by FileLog before writing — the filter takes priority over the constant. The directory must exist and be writable, otherwise FileLog falls back. Either approach moves
  logs out of the publicly accessible uploads directory.

File: src/Command/CommandSupervisor.php [hunk 4]
```diff
@@ -0,0 +1,345 @@
+<?php
+
+namespace Newspack\MigrationTools\Command;
+
+use Newspack\MigrationTools\Util\Log\MultiLog;
+use Psr\Log\LoggerInterface;
+use WP_CLI;
+use WP_CLI\ExitException;
+
```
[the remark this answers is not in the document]
eddiesshop: WP_CLI::halt() is used when failures exceed the established thresholds. That function throws an `ExitException`. So that's why this remains

src/Command/CommandSupervisor.php
```diff
@@ -0,0 +1,345 @@
...
+	 *
+	 * @param array $args Positional arguments.
+	 * @param array $assoc_args Associative arguments.
+	 *
+	 * @throws ExitException If an error occurs during command execution.
+	 */
+	public function cmd_supervise( array $args, array $assoc_args ): void {
+		$command              = $assoc_args['command'];
+		$max_fail_retries     = intval( $assoc_args['max-fail-retries'] ?? 3 );
+		$max_success_retries  = intval( $assoc_args['max-success-retries'] ?? 10 );
```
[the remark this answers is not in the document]
eddiesshop: Added a new test to cover usage of the new command.

src/Command/CommandSupervisor.php
```diff
@@ -0,0 +1,345 @@
...
+	 * @return string[] Plugin slugs to pass to --skip-plugins.
+	 */
+	private function get_plugins_to_skip( string $active_plugins_arg ): array {
+		// Build the whitelist of plugins that should remain active.
+		$whitelist = [ 'newspack-custom-content-migrator' ];
+		if ( ! empty( $active_plugins_arg ) ) {
+			$whitelist = array_merge(
+				$whitelist,
+				array_map( 'trim', explode( ',', $active_plugins_arg ) )
+			);
```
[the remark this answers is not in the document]
eddiesshop: Hardcode has been removed and replaced with a dynamic mechanism to obtain the plugin this tool is being invoked from.

src/Command/CommandSupervisor.php
```diff
@@ -0,0 +1,345 @@
...
+	 * @param array $assoc_args Associative arguments.
+	 *
+	 * @throws ExitException If an error occurs during command execution.
+	 */
+	public function cmd_supervise( array $args, array $assoc_args ): void {
+		$command              = $assoc_args['command'];
+		$max_fail_retries     = intval( $assoc_args['max-fail-retries'] ?? 3 );
+		$max_success_retries  = intval( $assoc_args['max-success-retries'] ?? 10 );
+		$retry_delay          = intval( $assoc_args['retry-delay'] ?? 5 );
+		$restart_on_success   = isset( $assoc_args['restart-on-success'] );
```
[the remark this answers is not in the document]
eddiesshop: These lines have been surrounded by `abs()` function calls now.

src/Command/CommandSupervisor.php
```diff
@@ -0,0 +1,426 @@
...
+					$this->logger->error( $process->stderr );
+				}
+				$this->logger->warning( sprintf( 'Command failed with exit code %d on attempt #%d (consecutive failures %d/%d, failure %d/%d).', $process->return_code, $attempt, $consecutive_fail_count, $max_consecutive_fail_retries, $total_fail_count, $max_fail_retries ) );
+
+				// Check whether we've exhausted failure retries.
+				if ( $total_fail_count >= $max_fail_retries ) {
+					$this->logger->warning( sprintf( 'Max fail retries (%d) reached. Stopping.', $max_fail_retries ) );
+					break;
+				}
+			}
```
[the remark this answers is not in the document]
eddiesshop: This has been cleaned up so that when $consecutive_fail_count is greater than $max_consecutive_fail_retries, a break is executed, stopping the script from proceeding further.

tests/Command/TestCommandSupervisor.php
```diff
@@ -0,0 +1,655 @@
...
+	 *
+	 * @param array $plugins The active plugins array to use.
+	 */
+	private function set_active_plugins( array $plugins ): void {
+		add_filter(
+			'pre_option_active_plugins',
+			function () use ( $plugins ) {
+				return $plugins;
+			}
+		);
```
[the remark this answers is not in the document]
eddiesshop: A new function to remove any filters added during the tests has bene implemented.

tests/Command/TestCommandSupervisor.php
```diff
@@ -0,0 +1,655 @@
...
+		$captured_atts = null;
+		add_filter(
+			'pre_wp_mail',
+			function ( $no_value, $atts ) use ( &$captured_atts ) {
+				$captured_atts = $atts;
+				return true;
+			},
+			10,
+			2
+		);
```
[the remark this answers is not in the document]
eddiesshop: A new function to remove any filters added during the tests has bene implemented.

tests/Command/TestCommandSupervisor.php
```diff
@@ -0,0 +1,655 @@
...
```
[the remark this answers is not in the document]
eddiesshop: A new function to remove any filters added during the tests has bene implemented.

tests/Command/TestCommandSupervisor.php
```diff
@@ -0,0 +1,655 @@
...
```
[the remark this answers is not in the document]
eddiesshop: A new function to remove any filters added during the tests has bene implemented.

src/Command/CommandSupervisor.php
```diff
@@ -0,0 +1,426 @@
...
+			if ( ! is_email( $notify_email ) ) {
+				$this->logger->error( sprintf( 'Invalid notify-email address provided: %s', $notify_email ) );
+			} else {
+				$this->send_notification( $notify_email, $final_exit_code, $command, $attempt );
+			}
+		}
+
+		if ( $total_success_count >= $max_success_retries ) {
+			$this->logger->info( 'Command succeeded after max success retries.' );
+		} elseif ( $consecutive_fail_count >= $max_consecutive_fail_retries || $total_fail_count >= $max_fail_retries ) {
```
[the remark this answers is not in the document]
eddiesshop: This has been updated so that only exceeding any of the max fail counters yields a `WP_CLI::halt()`, to signify to the caller that this failed because the script surpassed their established thresholds.

src/Command/CommandSupervisor.php
```diff
@@ -0,0 +1,430 @@
...
+						],
+						[
+							'type'        => 'flag',
+							'name'        => 'restart-on-success',
+							'description' => 'Also restart the command after a successful exit (exit code 0). Useful for batch-processing commands that need to run multiple passes. Requires --completion-criteria or --max-success-retries.',
+							'optional'    => true,
+						],
+						[
+							'type'        => 'assoc',
+							'name'        => 'completion-criteria',
```
kariae, line 82: Can we use exit codes instead, for this?

src/Command/CommandSupervisor.php
```diff
@@ -0,0 +1,430 @@
...
+						[
+							'type'        => 'assoc',
+							'name'        => 'max-success-retries',
+							'description' => 'Maximum number of successful runs before stopping. Only used with --restart-on-success when --completion-criteria is not set.',
+							'optional'    => true,
+							'default'     => 10,
+						],
+						[
+							'type'        => 'assoc',
+							'name'        => 'active-plugins',
```
kariae, line 95: Let's have a list of defaults that we use in most migrations:

- Co-Authors Plus
- Simple Local Avatar
- Yoast
- Newspack (for the brands migrations)
- ...

src/Command/CommandSupervisor.php
```diff
@@ -0,0 +1,430 @@
...
+		$retry_delay                  = abs( intval( $assoc_args['retry-delay'] ?? 5 ) );
+		$restart_on_success           = isset( $assoc_args['restart-on-success'] );
+		$completion_criteria          = $assoc_args['completion-criteria'] ?? null;
+		$active_plugins_arg           = $assoc_args['active-plugins'] ?? '';
+		$notify_email                 = $assoc_args['notify-email'] ?? null;
+
+		$this->logger = MultiLog::get_cli_and_file_logger( 'CommandSupervisor' );
+
+		// Strip the "wp" prefix if the caller included it.
+		$command = preg_replace( '/^\S*wp\s+/', '', $command );
```
kariae: The regex to remove `wp` is broad (removing all the strings that end with `wp`). We can make it to remove `wp` only.

src/Command/CommandSupervisor.php
```diff
@@ -0,0 +1,430 @@
...
+	 */
+	private function inject_wp_cli_global_flags( string $command, string $active_plugins_arg ): string {
+		// Check if the caller already included --skip-* flags in the command.
+		$has_skip_themes  = (bool) preg_match( '/--skip-themes\b/', $command );
+		$has_skip_plugins = (bool) preg_match( '/--skip-plugins\b/', $command );
+
+		$flags = [];
+
+		if ( $has_skip_themes ) {
+			$this->logger->info( '--skip-themes already present in command; skipping.' );
```
kariae: We can skip logging this, IMO.

src/Command/CommandSupervisor.php
```diff
@@ -0,0 +1,430 @@
...
+		} else {
+			$flags[] = '--skip-themes';
+		}
+
+		if ( $has_skip_plugins ) {
+			$this->logger->info( '--skip-plugins already present in command; skipping.' );
```
kariae: Same here.

kariae (changes requested):
Works like a charm. I left a few suggestions

src/Command/CommandSupervisor.php
eddiesshop: Copy 👍

src/Command/CommandSupervisor.php
eddiesshop: Copy 👍

src/Command/CommandSupervisor.php
eddiesshop: The main reason for `--completion-criteria` is the `--restart-on-success` use case. When the supervisor is intentionally restarting commands that exit 0 (for batch processing), exit codes alone can't distinguish between "succeeded with more work to do" and "succeeded and fully done."

We'd need custom exit codes (e.g. exit 2 = "complete") across migration commands to achieve the same thing, which would add coupling between the supervisor and the commands it supervises. String matching keeps it decoupled — it works with existing commands out of the box since many already output something like "No more posts to process."

Open to other approaches if you have something different in mind though!

src/Command/CommandSupervisor.php
eddiesshop: Updated so that it isn't so broad. Will only remove `wp` at the beginning of a command string.

src/Command/CommandSupervisor.php
eddiesshop: Forgot to leave a comment on here. I updated this so that it implements a constant + filter combo for obtaining a list of plugins that should be whitelisted. This will help keep it agnostic while allowing us the flexibility to default the values to the plugins we typically use for our migrations.

src/Command/CommandSupervisor.php
kariae: I understand the concern about coupling, but I think string matching on output is actually more fragile than exit codes. Output format isn't a stable contract, exit codes are.

We can define a completion exit code (e.g., `NMT::EXIT_DONE = 2`) and have commands call WP_CLI::halt(2) when there's no more work. The supervisor then treats exit 0 as "success, restart" and exit 2 as "done, stop" when `--restart-on-success` is set.

This is a one-line change per batch command, gives us an explicit and reliable signal, and avoids the risk of false positives from string matching (e.g., the completion string appearing in error output or migrated content).

If we want to keep `--completion-criteria` as a fallback for edge cases, that's fine, but exit codes should be the primary mechanism.

src/Command/CommandSupervisor.php
kariae: I like the flexibility of the constant + filter approach, but I think for our use case we can keep it simpler. Since we're the only consumers of NMT/NCCM, and our migrations pretty much always depend on the same plugins (Co-Authors Plus, Simple Local Avatars, Yoast, Newspack), I'd lean toward just hardcoding those as defaults in the class.

These are lightweight plugins, so keeping them active won't meaningfully affect performance, the real win from --skip-plugins comes from skipping heavy plugins. And it saves us from having to configure defaults before every migration run.

kariae: We can still use --active-plugins for anything extra, and if we ever need runtime configuration we can always add the filter later. What do you think?

src/Command/CommandSupervisor.php
eddiesshop: Exit codes are the primary mechanism now. I left `--completion-criteria` only as a fallback.

src/Command/CommandSupervisor.php
eddiesshop: The plugins above have been hardcoded. `--active-plugins` are kept active, in addition to the default plugins.

PR #169, opened 2026-04-03
Fix PHPUnit tests for no internet

ronchambers:
## Problem

Our current PHPUnit tests are dependent on having a live internet connection to run.  Most likely we'll always be connected to the internet, but in an effort for our tests to be true "unit tests", they should not depend on external connections that could fail without our control.

Luckily it's just a few image references, a PDF, and an mp3 file that need to be saved to the `fixtures` folder to use local files instead of fetching from the internet.

## Solution

1. Change references for external images, etc, to use local fixtures instead.  Local fixture files were copied from https://github.com/WordPress/wordpress-develop/.
2. Clean up images, etc, from `tmp` and `uploads` folders for any subsequent tests.  (PHPUnit will clear the db after each test, but it doesn't automatically clean up the disk).
3. Minor tweaks: fixed a broken PDF reference and a couple redundant lines of code.

## Test

1. Don't pull this PR yet.  Just try to run our current PHPUnit tests without internet access.  Notice the tests will fail.
2. Now, pull this PR (don't forget to turn your internet back on to do the pull).  Turn your internet off, then run PHPUnit tests.  Notice the tests will now work fine without internet.

## Future

* You may notice a left over file named `ghost_kg_elements.jsonl` after running the tests. This is it's own issue that I'll work on fixing later...please ignore.

* Once this PR is merged, I'll fix a different pending PR I worked on that did multiple internet fetches so that it uses this new methodology instead.  Thanks.


ronchambers:
@kariae @iuravic @naxoc @jorostoyanov - Please give this a quick look if you can.  It's just some clean up I stumbled upon while working on a different PR.  Let me know if you have questions/concerns.  Thanks!

kariae:
@ronchambers this works as described 👍 One question though, if we don't have internet it won't be impossible to install the WordPress test site `./bin/install-wp-tests.sh` at first, no?

ronchambers:
@kariae - correct, without internet everything breaks anyway 😉  _But_ ... when it comes to the individual tests themselves, it would be nice to try to limit the external resources needed for them to run.  If a developer already has the testing framework installed, or something like Github actions is reusing an already installed testing framework, then trying to minimize additional external requirements within the individual tests will speed things up.

The biggest case is our external requirement for [https://dummyimage.com](https://dummyimage.com)

We don't control this domain.  If it gets hacked, becomes really slow, etc, it could make our tests fail.  This PR makes it so we use files/images from local disk (much faster).  

Another is our test MP3 is being fetched from an external provider: [https://cdnpk.net](https://cdnpk.net)

This PR solves these cases...and for me personally, I have a pending PR that I'd like to use this new methodology of not needing external resources.

If for some reason in the future a test really can only run by fetching something external, we should try to fetch only from domains we control.

kariae:
Makes more sense now, thank you! Feel free to merge 👍 I don't like relying on resources from outside of the repo for the tests!

